PR #3, opened 2019-05-01
Fix CONFIG_x86 check

oddstr13:
It seems the x86_64 check got broken with the addition of arm64
(Changing `HOST_ARCH` -> `BUILD_ARCH`)
```bash
$ arch
x86_64
$ dpkg-architecture -q DEB_HOST_MULTIARCH
x86_64-linux-gnu
```


oddstr13:
_I was lazy, so I did a web edit - aka not tested._

joshuaboniface (approved):
Good catch.

PR #9, opened 2019-10-02
Add patch fixing up segment muxer

JustAMan:
This should close https://github.com/jellyfin/jellyfin/issues/1694.
Also it provides us the means to add more patches in the future if we wanted to.

This patch was sent to upstream (so we won't have to maintain our custom `ffmpeg` forever), initial thread is here: http://ffmpeg.org/pipermail/ffmpeg-devel/2019-October/250920.html

PR #10, opened 2019-10-02
 Add patch fixing up segment muxer

### LINKED ISSUE #1694: Video playback will end prematurely
**Describe the bug**
I haven't found any specific conditions yet, it seems to happen for any video file during playback. I am running the latest nightly and ffmpeg 4.1 from ubuntu. Not sure if the log below is the cause but it's the only suspicious entry I saw. If another video file is queued it will skip to that video.

**Logs**
`HLS Error: Type: mediaError Details: bufferStalledError Fatal: false`

joshuaboniface:
Closes #9 
Fixes jellyfin/jellyfin#1694

Re-implementation of #9 as a Debian source patch (rather than a manual patch). Also includes the porting of this patch to the Docker ffmpeg build.

Manual patch required until the relevant patch is accepted upstream. Tracking at http://ffmpeg.org/pipermail/ffmpeg-devel/2019-October/250920.html

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -316,8 +316,14 @@ RUN  \
         curl -sLO https://ffmpeg.org/releases/ffmpeg-${FFMPEG_VERSION}.tar.bz2 && \
         tar -jx --strip-components=1 -f ffmpeg-${FFMPEG_VERSION}.tar.bz2
 
+COPY debian/patches/*.patch /tmp/ffmpeg/
+
 RUN \
         DIR=/tmp/ffmpeg && mkdir -p ${DIR} && cd ${DIR} && \
+        for patch in *.patch; do \
+            filename="$( grep '+++' ${patch} | awk '{ print $2 }' | sed 's|jellyfin-ffmpeg/||' )"; \
+            patch --verbose ${filename} ${patch}; \
+        done && \
         ./configure \
         --disable-debug \
         --disable-doc \
--- a/debian/patches/0001_fix-segment-muxer.patch
+++ b/debian/patches/0001_fix-segment-muxer.patch
@@ -1,0 +1,37 @@
+Index: jellyfin-ffmpeg/libavformat/segment.c
+===================================================================
+--- jellyfin-ffmpeg.orig/libavformat/segment.c	2019-10-02 13:05:57.119005772 -0400
++++ jellyfin-ffmpeg/libavformat/segment.c	2019-10-02 13:06:11.367388679 -0400
+@@ -87,6 +87,7 @@
+     int64_t last_val;      ///< remember last time for wrap around detection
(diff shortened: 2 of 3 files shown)


andrewrabert (changes requested):
nit - the **commit messages** aren't clear why the patch is needed. Also lacks a link to the upstream contribution. 

```
Patch segment muxer in ffmpeg

This is necessary to fix xyz until it is fixed upstream.
Upstream contribution can be tracked at.
```

andrewrabert:
@joshuaboniface commit message, not pull request description. Pull request messages aren't included in the actual git repository AFAIK.

Use `git commit --amend` and/or `git rebase` - then force push to your branch.

joshuaboniface:
~~@nvllsvm Yes they are, the merge commit includes the full text of the PR description.~~

Or, not, I thought it did. That's stupid.

andrewrabert (approved):

docker/Dockerfile
```diff
@@ -316,8 +316,14 @@ RUN  \
         curl -sLO https://ffmpeg.org/releases/ffmpeg-${FFMPEG_VERSION}.tar.bz2 && \
         tar -jx --strip-components=1 -f ffmpeg-${FFMPEG_VERSION}.tar.bz2
 
+COPY debian/patches/*.patch /tmp/ffmpeg/
+
 RUN \
         DIR=/tmp/ffmpeg && mkdir -p ${DIR} && cd ${DIR} && \
+        for patch in *.patch; do \
+            filename="$( grep '+++' ${patch} | awk '{ print $2 }' | sed 's|jellyfin-ffmpeg/||' )"; \
+            patch --verbose ${filename} ${patch}; \
```
JustAMan, line 325: Heh, why not just `patch -p1 < "$patch"`?

JustAMan (approved):

docker/Dockerfile
joshuaboniface: For whatever reason, the patch command wasn't liking the Debian patch format, mostly because the actual path to the file was different (we're one level lower at this stage). This makes it super explicit.

JustAMan: But that is what `-pN` does - it drops first `N` dirs in a path before applying the patch...

JustAMan: Also your method would fail for multi-file patches...

joshuaboniface: Oh really? OK, yea, that's nicer. I went through the man page quickly and didn't see the obvious solution. I guess that's an improvement to be made if we implement any more patches ;-) Hopefully we don't and this is redundant before then anyways.

JustAMan:
For future reference, here's better upstream tracker URL: https://patchwork.ffmpeg.org/patch/15492/

PR #13, opened 2019-10-22
Support Ubuntu Eoan

joshuaboniface:
Adds Ubuntu Eoan (19.10) support to jellyfin-ffmpeg.

### DIFF
--- a/build
+++ b/build
@@ -10,6 +10,7 @@ usage() {
     echo -e " * bionic"
     echo -e " * cosmic"
     echo -e " * disco"
+    echo -e " * eoan"
 }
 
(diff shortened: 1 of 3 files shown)


build
```diff
@@ -10,6 +10,7 @@ usage() {
     echo -e " * bionic"
     echo -e " * cosmic"
     echo -e " * disco"
+    echo 0e " * eoan"
```
bugfixin: Should this be -e rather than 0e?

dkanada (approved):

build
joshuaboniface: LOL yup. Thanks for catching my typo!

PR #14, opened 2019-11-14
Add MMAL for armhf on Raspberry Pi

joshuaboniface:
Includes a download of the Raspberry Pi headers and GPU driver in the
armhf build process, in order to properly enable MMAL cross-compiling.
This is OK license wise both due to the licenses of these headers (LGPL
and MIT mostly) and since these are not included in the final binary
packages.


ferferga:
Finally! Thank you very much for helping me in this! 👏👏

PR #19, opened 2020-01-19
Initial steps towards ARM support

andrewrabert:
- introduces architecture-dependent branches to the Dockerfile
- upgrade x265 to 3.2.1


JustAMan (approved):
What makes it WIP?

andrewrabert:
ARM still doesn't build correctly. It fails on the ffmpeg compilation.

However, there's no reason this couldn't merge as x86 would still build.

PR #21, opened 2020-02-15
Support Debian Bullseye (testing)

joshuaboniface:


Artiume (approved):
Cool, I'll start testing docker arm.

What is gcc version?

joshuaboniface:
@Artiume I picked 8 since it was available, but that is flexible.

Artiume:
I'll read through the build scripts. I wasn't sure what it was for, but I think I have an understanding now.

andrewrabert (approved):

build.yaml
```diff
@@ -1,14 +1,17 @@
 ---
 # We just wrap `build` so this is really it
 name: "jellyfin-ffmpeg"
-version: "4.2.1-2"
+version: "4.2.1-4"
```
andrewrabert, line 4: curious - where did 3 go?

joshuaboniface: It existed, but we never updated the build.yaml - doesn't seem to have mattered though.

PR #22, opened 2020-02-19
Check for dependencies on start

joshuaboniface:
Ensures all the dependencies exists before kicking off a build.

### DIFF
--- a/build
+++ b/build
@@ -81,6 +81,11 @@ esac
 set -o xtrace
 set -o errexit
 
+# Check for dependencies
+for dep in docker make mmv; do
+    which $dep || { echo "The command 'make' is required."; exit 1; }
+done
+
 image_name="jellyfin-ffmpeg-build-${cli_release}"
 package_temporary_dir="$( mktemp -d )"
 current_user="$( whoami )"
--- a/build
+++ b/build
@@ -83,7 +83,7 @@ set -o errexit
 
 # Check for dependencies
 for dep in docker make mmv; do
-    which ${dep} || { echo "The command '${dep}' is required."; exit 1; }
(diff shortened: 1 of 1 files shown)


Artiume (approved):

build
```diff
@@ -81,6 +81,11 @@ esac
 set -o xtrace
 set -o errexit
 
+# Check for dependencies
+for dep in docker make mmv; do
+    which $dep || { echo "The command 'make' is required."; exit 1; }
```
JustAMan:
```suggestion
    which $dep || { echo "The command '$dep' is required."; exit 1; }
```

JustAMan (changes requested):

build
```diff
@@ -81,6 +81,11 @@ esac
 set -o xtrace
 set -o errexit
 
+# Check for dependencies
+for dep in docker make mmv; do
+    which ${dep} || { echo "The command '${dep}' is required."; exit 1; }
```
andrewrabert: nit - `command -v` is a builtin and thus should be "faster"

andrewrabert (approved):

Artiume (approved):

build
joshuaboniface: Interesting, didn't know about that one!

build
```diff
@@ -83,7 +83,7 @@ set -o errexit
 
 # Check for dependencies
 for dep in docker make mmv; do
-    which ${dep} || { echo "The command '${dep}' is required."; exit 1; }
+    command -v ${dep} &>/dev/null || { echo "The command '${dep}' is required."; exit 1; }
```
JustAMan, line 86: should be `command -v "${dep}"` in case `dep` contains a space. I know it most likely wouldn't, so it's a nitpicking... but still %)

PR #28, opened 2020-02-25
Update AMF encoder to support Linux

nyanmisaka:
Update some AMF files to [support linux via vulkan](https://github.com/FFmpeg/FFmpeg/commit/f8ad2ddd7a51df5d6e768ba2a6f65ee64b94e29b#diff-3adfff78d6c720bbaa2e75bd85524380) and fix minor bugs.

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-#NVCH_VERSION="8.2.15.8-dmo1"
-NVCH_VERSION="9.0.18.3-dmo1"
-
 # Builds the DEB inside the Docker container
 
 set -o errexit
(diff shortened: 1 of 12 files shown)


docker-build.sh
```diff
@@ -114,10 +114,12 @@ popd
 
 # Download and setup AMD AMF headers from AMD official github repo
 # https://www.ffmpeg.org/general.html#AMD-AMF_002fVCE
-wget -O amf_headers.zip https://github.com/GPUOpen-LibrariesAndSDKs/AMF/archive/master.zip
-unzip amf_headers.zip -d amf_headers
-cd "amf_headers/AMF-master/amf/public/include/"
-mkdir -p "/usr/include/AMF/" && cp -r * "/usr/include/AMF/"
+apt-get update
+yes | apt-get install subversion
+svn checkout https://github.com/GPUOpen-LibrariesAndSDKs/AMF/trunk/amf/public/include
```
JustAMan, line 119: but why? what was wrong with downloading `master`?

libavcodec/amfenc.c
```diff
@@ -311,8 +312,20 @@ static int amf_init_context(AVCodecContext *avctx)
             if (res == AMF_OK) {
                 av_log(avctx, AV_LOG_VERBOSE, "AMF initialisation succeeded via D3D9.\n");
             } else {
-                av_log(avctx, AV_LOG_ERROR, "AMF initialisation failed via D3D9: error %d.\n", res);
-                return AVERROR(ENOSYS);
+                AMFGuid guid = IID_AMFContext1();
```
JustAMan: how we ended up hosting source files of ffmpeg?..

docker-build.sh
nyanmisaka: The header file required for compilation is only 433 KB, while the entire master.zip is 115 MB. Many of these files are unnecessary to download.

JustAMan: LOL, okay :)

libavcodec/amfenc.c
nyanmisaka: These changes should be available in FFmpeg 4.3 but I don't know when it will be released and when jellyfin will follow up on the upstream version.

JustAMan: Why don't you include it as a patch, so normal tree stays clean? See here: https://github.com/jellyfin/jellyfin-ffmpeg/tree/jellyfin/debian/patches

joshuaboniface:
Yes, please use a Debian source patch, rather than editing any upstream code. The upstream code is overwritten wholesale when there is a new version.

nyanmisaka:
Ok. I see, I will try to solve this problem with patch.

joshuaboniface:
The Quilt guide should be useful for you here: https://wiki.debian.org/UsingQuilt#Making_a_new_patch

It's basically a similar process, except using Quilt to capture your edits and then storing them in a `.patch` file, which would then be applied at build-time.

You could probably soft-reset your changes, stash them, do the `quilt add` steps, then stash apply and `quilt refresh` to avoid redoing all your changes ;-)

libavcodec/amfenc.c
nyanmisaka: I have made a patch to solve it : )

joshuaboniface (approved):

JustAMan (approved):
Hope that builds :)

Issue #29, opened 2020-02-25
V4L2 issues with ffmpeg


aptalca:
Hi guys,

Just want to report that I've been doing some tests with hw encode on an rpi4 via both openmax and v4l2. Openmax seems to work as expected, without issues (encode only).

V4L2 on the other hand results in some issues and errors (also encode only).

The encoded video appears in shades of green, no other colors.

Here's a full ffmpeg log output: https://pastebin.com/h6Awbv9X
Here are the errors that occur:
```
[h264_v4l2m2m @ 0x2fc65e0] Failed to set number of B-frames
    Last message repeated 1 times
[h264_v4l2m2m @ 0x2fc65e0] Failed to set header mode
[h264_v4l2m2m @ 0x2fc65e0] Failed to set gop size
[h264_v4l2m2m @ 0x2fc65e0] h264 profile not found
[h264_v4l2m2m @ 0x2fc65e0] Encoder adjusted: qmin (0), qmax (51)
[h264_v4l2m2m @ 0x2fc65e0] Failed to set minimum video quantizer scale
[h264_v4l2m2m @ 0x2fc65e0] Failed to set maximum video quantizer scale
```

The helpful folks over at LibreELEC pointed me in the right direction (especially 6by9 from raspberry pi). The shades of green is due to jellyfin's ffmpeg missing [this commit](https://github.com/FFmpeg/FFmpeg/commit/d61cf1b1ebc2477749d7d7825a072400ed24af9f#diff-bec5e20dbf17e1c989823fa841c2f8b3) which is merged to master but is not in 4.2

I'll let you decide whether you want to patch it for the arm/aarch64 builds or not. But I thought you should know.

Thanks.

Artiume:
Hey there, I made a PR for this. Unfortunately, your ffmpeg logs are gone, can i get them again?

aptalca:
@Artiume 
Here's a fresh log with the latest jellyfin stable, trying v4l2 for encode (getting green screen): https://pastebin.com/JSRzjXG6

By the way, in the previous version of jellyfin, openmax was working fine, but in the latest it is no longer working. I'm looking into it and will post logs in a bit.

aptalca:
When I select openmax (only for encode, the decode ones are unchecked), playback never starts and results in a pop up error: 
```
Playback Error
There was an issue with the client profiling and the server isn't sending a compatible media format.
```
Here's the log: https://pastebin.com/0hAFiwdB

[...]

PR #31, opened 2020-03-08
Rpi Support - Add handling for NV21 and YUV420P

Artiume:
untested.

https://github.com/jellyfin/jellyfin-ffmpeg/issues/29

https://github.com/FFmpeg/FFmpeg/commit/d61cf1b1ebc2477749d7d7825a072400ed24af9f?diff=unified

### DIFF
--- a/libavcodec/v4l2_buffers.c
+++ b/libavcodec/v4l2_buffers.c
@@ -321,11 +321,20 @@ int ff_v4l2_buffer_buf_to_avframe(AVFrame *frame, V4L2Buffer *avbuf)
     /* 1.1 fixup special cases */
     switch (avbuf->context->av_pix_fmt) {
     case AV_PIX_FMT_NV12:
+    case AV_PIX_FMT_NV21:
         if (avbuf->num_planes > 1)
             break;
         frame->linesize[1] = avbuf->plane_info[0].bytesperline;
         frame->data[1] = frame->buf[0]->data + avbuf->plane_info[0].bytesperline * avbuf->context->format.fmt.pix_mp.height;
         break;
+   case AV_PIX_FMT_YUV420P: 
+        if (avbuf->num_planes > 1)
+            break;
+        frame->linesize[1] = avbuf->plane_info[0].bytesperline >> 1;
+        frame->linesize[2] = avbuf->plane_info[0].bytesperline >> 1;
+        frame->data[1] = frame->buf[0]->data + avbuf->plane_info[0].bytesperline * avbuf->context->format.fmt.pix_mp.height;
+        frame->data[2] = frame->data[1] + ((avbuf->plane_info[0].bytesperline * avbuf->context->format.fmt.pix_mp.height) >> 2);
+            break;
     default:
         break;
     }


JustAMan (commented):
We decided to not modify the source but supply patches, don't we?

JustAMan:
Also it just came to my mind - @nvllsvm, do we apply patches when building a dockerizef version of our ffmpeg?

andrewrabert:
We don't modify the source code directly as we don't want to maintain a fork. I find working with patch files easier in this situation.

they are applied in the Dockerfile.

Artiume:
So how would I apply this patch exactly?

Since this patch is applied in master, should the patch itself come from their master?

ferferga:
Can't we cherrypick the [original commit](https://github.com/FFmpeg/FFmpeg/commit/d61cf1b1ebc2477749d7d7825a072400ed24af9f#diff-bec5e20dbf17e1c989823fa841c2f8b3) that @Artiume is trying to merge? He's not making changes that could make the maintenance more dificult, what he is doing it's already in master in the official repository, but not merged to 4.2.

Artiume:
verified build, i dont have a sample file to test the fix with though

andrewrabert:
IMO - lets wait until upstream releases a stable version with this patch.

JustAMan:
We already have some backports in our patches, so it shouldn't be that big of a deal. Though I'm still worried about breaking `yuv420p` colorspace...

joshuaboniface:
@Artiume Please reimplement as a Debian package patch (though this wouldn't help the Docker build) or close it out until upstream is consistent.

Artiume:
I'm not sure how to package it. What can we do to fix docker? @nyanmisaka do you know if the version bump is ffmpeg would fix this or is still master only?

nyanmisaka:
> I'm not sure how to package it. What can we do to fix docker? @nyanmisaka do you know if the version bump is ffmpeg would fix this or is still master only?

It is master only as I cant find the patch in the latest stable 4.2.2.

If you switch the ffmpeg in Dockerfile.arm/arm64 to use jellyfin-ffmpeg, and then apply this patch, it should solve the problem.

This guide should be useful when making a deb patch.
https://github.com/jellyfin/jellyfin-ffmpeg/pull/28#issuecomment-591497304

JustAMan:
> If you switch the ffmpeg in Dockerfile.arm/arm64 to use jellyfin-ffmpeg, and then apply this patch, it should solve the problem.

We should really do this ASAP, or we risk getting different ffmpeg builds in Docker and outside in the future if someone forgets about this duality.

Artiume:
> This guide should be useful when making a deb patch.
#28 (comment)

I would need assistance in making this happen. What would be the difference between Quilt and my PR? Once the change from upstream is ready, we just remove the patch file and it removes the change? Thus we become aligned with upstream again?

Artiume:
Ffmpeg 4.2.3 is out.
https://ffmpeg.org/download.html#releases

dfunkt:
FFmpeg 4.3 is out and it seems to include the commit from this PR.

joshuaboniface:
With 4.3 out, and 4.3 being merged in other PRs, I'm going to close this one.

PR #36, opened 2020-04-19
Add Ubuntu Focal support

joshuaboniface:


nyanmisaka:
I have added support for Focal in #33 but forgot to mention it in changlog, so these two may conflict. In addition, the default gcc version in Focal is [9](https://packages.ubuntu.com/focal/amd64/gcc), why not use it ;)

PR #43, opened 2020-06-18
Add AV1 decoding support with dav1d

### LINKED ISSUE #39: av1 (via dav1d) support
Add support for dav1d (and maybe libaom if it's needed), a av1 decoder ([build guide](https://xtechbiz.com/ffmpeg-dav1d-static-build-ubuntu-19-04/)).

This should enables ffmpeg to handle av1 transcoding correctly.

This seems to require ffmpeg 4.3 or 4.2 not sure, so it might have to be tested.

RazeLighter777:
Adds dav1d library support to jellyfin-ffmpeg. This will allow jellyfin-ffmpeg to transcode AV1 files to other codecs such as H.264. This Fixes #39. I'm not experienced with debian packaging.

Note that this pull request does NOT allow transcoding to AV1, encoders for AV1 are not in a viable state currently. This allows playback of AV1 files by transcoding them from AV1 to another format.

(EX AV1 file -> dav1d -> x264 -> Jellyfin client)

dav1d is pulled from upstream and built statically (EDIT: Now dynamically) against FFmpeg. I chose to do this because there are dav1d is relatively new and not packaged for many distributions.

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -25,7 +25,7 @@ prepare_hwa_amd64() {
     pushd dav1d
     mkdir build
     pushd build
-    meson --prefix=${TARGET_DIR} ..
+    meson -Ddefault_library=static --prefix=${TARGET_DIR} ..
     ninja
     popd
     popd
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -25,8 +25,14 @@ prepare_hwa_amd64() {
     pushd dav1d
     mkdir build
     pushd build
-    meson -Ddefault_library=static --prefix=${TARGET_DIR} ..
+    meson -Ddefault_library=shared --prefix=${TARGET_DIR} ..
     ninja
+    meson install
+    echo ${LD_LIBRARY_PATH}
+    echo $(find ${TARGET_DIR}/lib/x86_64-linux-gnu)
+    cp ${TARGET_DIR}/lib/x86_64-linux-gnu/pkgconfig/dav1d.pc  /usr/lib/pkgconfig/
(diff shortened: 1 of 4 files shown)


RazeLighter777:
Currently need to fix the static build, because all dependencies like libx265 and others were unintentionally linked statically. I assume this is not desired behavior.

RazeLighter777:
Upon further review, I decided to use a shared library instead to prevent muddling the build configuration. Dav1d is compiled and included as a shared library with jellyfin-ffmpeg.

RazeLighter777:
@joshuaboniface 
@nvllsvm 
My changes are ready for review. To build, use command `./build [distro] [arch]` and the packages will be placed in `../bin/jellyfin-ffmpeg-whatever.deb`

docker-build.sh
```diff
@@ -19,6 +19,30 @@ prepare_hwa_amd64() {
     popd
     popd
 
+    # Download and install dav1d
```
nyanmisaka, line 22: `prepare_hwa_amd64()` only prepares files for x86. If so, `--enable-libdav1d` should also be placed in `CONFIG_x86`. Or put these code in a place behind to suit arm64 and armhf.

nyanmisaka (changes requested):

docker-build.sh
RazeLighter777: changes addressed. I moved the build option to x86 only.

RazeLighter777:
I am at boot camp and am unable to work on this pull request for the next month.

joshuaboniface (approved):
So far this looks good to me, I see no issues with it from a license or build standpoint, and it seems to build fine.

PR #41, opened 2020-06-16
Update docker ffmpeg to 4.3

Artiume:
Have not checked the codecs to see if any of them need to be updated.

PR #48, opened 2020-07-19
Bump to 4.3.1 and fix some bugs in Docker image

joshuaboniface:
Replaces #41

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -75,11 +75,11 @@ RUN      buildDeps="autoconf \
                     libtool \
                     make \
                     nasm \
-		    ninja-build \
+                    ninja-build \
                     perl \
                     pkg-config \
                     python \
-		    python3-setuptools \
+                    python3-setuptools \
                     python3-pip \
                     libgnutls28-dev \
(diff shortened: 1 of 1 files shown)


joshuaboniface:
This does still fail. Despite the configuration for this not having changed, It's unable to find the nvidia headers when running the `./configure` step:

```
ERROR: cuvid requested, but not all dependencies are satisfied: ffnvcodec

If you think configure made a mistake, make sure you are using the latest
version from Git.  If the latest version fails, report the problem to the
ffmpeg-user@ffmpeg.org mailing list or IRC #ffmpeg on irc.freenode.net.
Include the log file "ffbuild/config.log" produced by configure as this will help
solve the problem.
```

I don't see a good reason why, and the sheer amount of time it takes to debug this is annoying.

At the moment, I don't believe we actually use this Dockerfile for anything, as our main Docker images are installing `jellyfin-ffmpeg` via the Debian packages instead of pulling from an FFfmpeg image. I would like that to change in the future, but I don't want to block the 10.6.0 release or 4.3.1 ffmpeg release on this.

Issue #52, opened 2020-10-17
Unmet libx264 dependency when building jellyfin docker based on bullseye


bzzbrr:
**Describe the bug**
Attempting to build jellyfin docker based on Debian Bullseye fails due to unmet libx264 dependency.  Presumably this is because libx264-159 is not a bullseye package (anymore?).  [libx264-160 is available in bullseye (version 2:0.160.3011+gitcde9a93-2)](https://packages.debian.org/search?suite=default&section=all&arch=any&searchon=names&keywords=libx264).

I assume this is the correct repository for this issue out of the jellyfin repos.  I understand bullseye might not be supported, but it works in every other respect.  Apologies for any mistakes as this is my first reporting.

**System (please complete the following information):**
 - OS: Ubuntu 20.04 amd64
 - Jellyfin Version: stable

**To Reproduce**
1. Git clone jellyfin-metapackages
2. Update Dockerfile references from buster to bullseye
3. Build container
```
user@host:/path/build$ cd jellyfin-metapackages/
user@host:/path/build/jellyfin-metapackages$ diff Dockerfile.amd64.original Dockerfile.amd64.bullseye
6c6
< FROM debian:buster-slim
---
> FROM debian:bullseye-slim
26c26
<  && echo 'deb [arch=amd64] https://repo.jellyfin.org/debian buster main' > /etc/apt/sources.list.d/jellyfin.list \
---
>  && echo 'deb [arch=amd64] https://repo.jellyfin.org/debian bullseye main' > /etc/apt/sources.list.d/jellyfin.list \

user@host:/path/build/jellyfin-metapackages$ docker build -f Dockerfile.amd64.bullseye -t bulltest/jellyfin .
```

**Logs**
```
deb [arch=amd64] https://repo.jellyfin.org/debian bullseye main
Hit:1 http://security.debian.org/debian-security bullseye-security InRelease
Hit:2 http://deb.debian.org/debian bullseye InRelease
Get:3 https://repo.jellyfin.org/debian bullseye InRelease [6639 B]
Hit:4 http://deb.debian.org/debian bullseye-updates InRelease
Get:5 https://sgp1.mirror.jellyfin.org/debian bullseye/main amd64 Packages [1694 B]
Fetched 8333 B in 12s (695 B/s)
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
openssl is already the newest version (1.1.1g-1).
openssl set to manually installed.
[...]

PR #53, opened 2020-10-24
Add support for Ubuntu Groovy 20.10

nyanmisaka:
*enable dav1d AV1 for armhf and arm64
*add libvpx encoder
*update libva to 2.9
*update gmmlib to 20.3.2
*update MediaSDK to 20.3

### DIFF
--- a/build
+++ b/build
@@ -29,11 +29,11 @@ case ${cli_release} in
     ;;
     'buster')
         release="debian:buster"
-        gcc_version="7"
+        gcc_version="8"
     ;;
     'bullseye')
         release="debian:bullseye"
-        gcc_version="8"
+        gcc_version="9"
     ;;
     'xenial')
         release="ubuntu:xenial"
--- a/build
+++ b/build
@@ -13,6 +13,7 @@ usage() {
     echo -e " * disco"
     echo -e " * eoan"
     echo -e " * focal"
+    echo -e " * groovy"
 }
 
 if [[ -z ${1} ]]; then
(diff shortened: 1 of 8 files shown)


nyanmisaka:
@joshuaboniface Rebuild jellyfin-ffmpeg should fix this issue https://github.com/jellyfin/jellyfin-ffmpeg/issues/52
```
Version: 4.3.1-2-bullseye
Depends: libx264-160 (>= 2:0.160.3011+gitcde9a93) ...
```

build
```diff
@@ -28,11 +29,11 @@ case ${cli_release} in
     ;;
     'buster')
         release="debian:buster"
-        gcc_version="7"
+        gcc_version="8"
```
joshuaboniface, line 32: Nothing should chamge for stable, why?

debian/rules
```diff
@@ -34,7 +34,9 @@ CONFIG := --prefix=${TARGET_DIR} \
 	--enable-libopus \
 	--enable-libtheora \
 	--enable-libvorbis \
+        --enable-libdav1d \
```
joshuaboniface: Spacing, but just a nitpick.

joshuaboniface (approved):
Looks good with a few comments. Will want to thoroughly test the build changes before tagging a release.

build
nyanmisaka: It has been updated to gcc 8.3 in Debian buster(stable).
https://packages.debian.org/buster/gcc

It throw me an 404 error when grabbing the gcc 7 package.

joshuaboniface: Thats weird but OK.

debian/rules
nyanmisaka: Done

joshuaboniface (approved):

PR #61, opened 2021-02-07
Add fdk-aac-free

knackebrot:
Fedora maintains a fork of libfdk-aac stripped of all AAC modes except for MPEG-2 AAC-LC, patents of which should now be expired. Their legal team greenlighted it, so it should be fine to redistribute such binaries. I'd like this encoder to replace the default LAME encoder that is in Jellyfin, because:
1. AAC is far more common on the Internet (less potential browser bugs)
2. better compression
3. Don't know if this is also in the mpegts muxer, but mp3 with variable bitrate couldn't seek accurately on some platforms. For Jellyfin, we could simply use fdk-aac VBR preset 5 or 4 instead of a fixed bitrate and save some bandwidth without sacrificing quality.

If this gets merged I'm planning to create a PR to the main Jellyfin repository to swap LAME for FDK-AAC.

Related links:
https://bugzilla.redhat.com/show_bug.cgi?id=1501522
https://frederickding.gitlab.io/HandBrake/

### DIFF
--- a/debian/patches/0035-remove-fdk-aac-from-nonfree.patch
+++ b/debian/patches/0035-remove-fdk-aac-from-nonfree.patch
@@ -1,0 +1,20 @@
+Index: jellyfin-ffmpeg/configure
+===================================================================
+--- jellyfin-ffmpeg.orig/configure
++++ jellyfin-ffmpeg/configure
+@@ -1730,7 +1730,6 @@ EXTERNAL_LIBRARY_GPL_LIST="
+ 
+ EXTERNAL_LIBRARY_NONFREE_LIST="
+     decklink
+-    libfdk_aac
+     openssl
+     libtls
+ "
+@@ -1771,6 +1770,7 @@ EXTERNAL_LIBRARY_LIST="
+     libdav1d
+     libdc1394
+     libdrm
++    libfdk_aac
+     libflite
+     libfontconfig
+     libfreetype
--- a/debian/patches/series
+++ b/debian/patches/series
@@ -32,3 +32,4 @@
 0032-add-async-support-for-qsv-vpp.patch
 0033-add-fixes-for-corrupted-hevc-vaapi-on-tgl.patch
 0034-add-webp-support-for-id3v2.patch
+0035-remove-fdk-aac-from-nonfree.patch
--- a/debian/rules
+++ b/debian/rules
@@ -42,6 +42,7 @@ CONFIG := --prefix=${TARGET_DIR} \
 	--enable-libx265 \
 	--enable-libzvbi \
 	--enable-libzimg \
+	--enable-libfdk-aac \
 
 CONFIG_ARM_COMMON := --toolchain=hardened \
 	--enable-cross-compile \
--- a/docker-build-win64.sh
+++ b/docker-build-win64.sh
@@ -437,6 +437,21 @@ meson install
 popd
 popd
 
+# FDK-AAC-STRIPPED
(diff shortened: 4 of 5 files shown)


nyanmisaka:
The default audio encoder in JF 10.7 has been changed to AAC encoder.
https://github.com/jellyfin/jellyfin-web/pull/2026

We also prefer to `libfdk_aac` encoder if it's available in a custom ffmpeg build.
However, no additional adjustments were made to its options.
https://github.com/jellyfin/jellyfin/pull/2715

Can we remove `--enable-nonfree` before adding that greenlighted FDK AAC library?

BTW since we do not force users to use jellyfin-ffmpeg, there is a high probability that there is no `libfdk_aac` encoder in a custom redistributable ffmpeg.

nyanmisaka:
Please use debian patch to modify ffmpeg sources.
https://github.com/jellyfin/jellyfin-ffmpeg/commit/60c4241ecc44576f4fe0c54f9ad39e4299f53639

knackebrot:
Sorry, I wasn't up-to-date with what was happening in the repos.
I've removed the library from the nonfree list and it's now building without --enable-nonfree.

knackebrot:
I've just found out that fdk-aac might not be worth it, at least at the bitrates we use. On my x86_64 server, it encodes at ~ 15x vs LAME's 93x. That's a 6x slowdown for audio that is indistinguishable, and at a similar bitrate. On my aarch64 phone it's 10x vs 55x. On devices with a weak FPU the difference might be slimmer, though. Still, LAME doesn't support surround sound, so here we must use aac.

joshuaboniface:
Do we still want this @nyanmisaka @knackebrot ?

knackebrot:
I'd like to have this although it's noticeably slower (maybe due to its reliance on integer math). It won't be an issue for most users, only those with lots of concurrent encodes, and I still stand behind my opinion that FFmpeg AAC is not very good.

knackebrot:
Hi @nyanmisaka @joshuaboniface
There were some changes in the upstream repository and now the abysmal performance i've been seeing is gone and it performs similarly to ffmpeg's aac encoder: 
```bash
ffmpeg -i "02 - Gouche.flac" -vn -c:a libfdk_aac -b:a 256k gouche.m4a

size=    8181kB time=00:04:20.27 bitrate= 257.5kbits/s speed=74.7x
```

```bash
ffmpeg -i "02 - Gouche.flac" -vn -c:a aac -b:a 256k gouche-ff.m4a

size=    8192kB time=00:04:20.27 bitrate= 257.9kbits/s speed=76.5x
```

I've built and tested it on win64 and ubuntu focal and would like to see it merged.

debian/patches/0035-remove-fdk-aac-from-nonfree.patch
```diff
@@ -0,0 +1,20 @@
+Index: jellyfin-ffmpeg2/configure
+===================================================================
+--- jellyfin-ffmpeg2.orig/configure
```
nyanmisaka: It'd be nice if you remove the extra 2.

nyanmisaka (commented):
LGTM. I want to do some testing before merging.

nyanmisaka:
IIRC in transcoding profile, aac encoder is preferred in jellyfin-web but mp3lame is preferred in other client such as android.

fdk_aac is preferred to aac [in the server side](https://github.com/jellyfin/jellyfin/blob/3fc8652c265fe0b9c6876f0b050aed9ef31b187c/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L489) too if it exists.

nyanmisaka (approved):

PR #63, opened 2021-03-25
Do not link against GnuTLS

strugee:
It's not needed for Jellyfin.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -13,8 +13,6 @@ Build-Depends:
  debhelper,
 # for build-profile support
  dpkg-dev,
-# --enable-gnutls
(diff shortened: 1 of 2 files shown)


strugee:
Note: I didn't actually test this since I assumed CI would. I'll do so later.

cvium:
Why is it not needed? We do occasionally feed ffmpeg https urls

strugee:
Oh really? When? I guess I just assumed we didn't.

cvium:
A .strm file may contain a link and then there's Live TV providers

strugee:
Ahhhh okay. Makes sense. Closing then, sorry for the noise!

PR #65, opened 2021-04-08
Enable link time optimization for ffmpeg

Bond-009:


docker-build.sh
```diff
@@ -26,7 +26,7 @@ prepare_extra_common() {
         ;;
     esac
     ./autogen.sh
-    ./configure --prefix=${TARGET_DIR} ${CROSS_OPT}
+    ./configure --prefix=${TARGET_DIR} --enable-lto ${CROSS_OPT}
```
nyanmisaka: This is only set for libzimg.
Shouldn't it be set here for ffmpeg? https://github.com/jellyfin/jellyfin-ffmpeg/blob/32815a2ad139198b21716682a343ef4d630e17f8/debian/rules#L94-L95

joshuaboniface:
What is the benefit/drawback/impact of this?

nyanmisaka:
> What is the benefit/drawback/impact of this?

https://johnysswlab.com/link-time-optimizations-new-way-to-do-compiler-optimizations/

> Compilation takes double the time with LTO enabled for both compilers. Linking time is huge with LTO for both compilers, on GCC it’s 86 times slower, on CLANG 48 times slower.

> This experiment has shown that, for ffmpeg at least, enabling LTO doesn’t bring expected speed ups.

Bond-009:
> What is the benefit/drawback/impact of this?

All my test are done on ffmpeg master commit 33db0cbfd08384d611370006a77675cc6b778d12
with gcc (GCC) 10.2.0 and clang version 11.1.0
on a Ryzen 7 2700 with 32GB of ram

| |GCC|GCC LTO|Clang|
|:-----:|:-----:|:-----:|:-----:|
|build time (s)|105|4,813|95|
|ffmpeg size|279,584|279,576|275,480|
|ffprobe size|168,040|155,712|180,344|

```
Benchmark #1: ./ffmpeg_gcc -threads 1 -i BigBuckBunny-AV1.webm -benchmark -f null -
  Time (mean ± σ):      1.188 s ±  0.015 s    [User: 1.175 s, System: 0.013 s]
  Range (min … max):    1.163 s …  1.212 s    10 runs
```

```
Benchmark #1: ./ffmpeg_gcc_lto -threads 1 -i BigBuckBunny-AV1.webm -benchmark -f null -
  Time (mean ± σ):      1.180 s ±  0.014 s    [User: 1.167 s, System: 0.012 s]
  Range (min … max):    1.156 s …  1.201 s    10 runs
```

```
Benchmark #1: ./ffmpeg_clang -threads 1 -i BigBuckBunny-AV1.webm -benchmark -f null -
  Time (mean ± σ):      1.165 s ±  0.013 s    [User: 1.156 s, System: 0.009 s]
  Range (min … max):    1.150 s …  1.188 s    10 runs
```

The build time difference on GitHub actions was an increase of 3-5min (on a 20min build)

I'd say it's worth it, even if the advantages are (currently) limited, they will only improve with newer compilers.

crobibero (approved):

nyanmisaka:
Don't forget to add it for windows build.

https://github.com/jellyfin/jellyfin-ffmpeg/blob/99b6c4711c8b00b0bef3231f78eb8334ba58ec6e/docker-build-win64.sh#L514

joshuaboniface (approved):

nyanmisaka (approved):
LTGM, even though it will slow down the build speed

PR #66, opened 2021-04-10
Add GitHub Actions workflow

daullmer:
- Add GitHub Action for building Linux Packages
- Change build script to allow for different .deb location


h1dden-da3m0n (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,38 @@
+name: Build Linux
+
+on:
+  push:
+    branches: [ jellyfin ]
```
ferferga:
```suggestion
    branches:
          - jellyfin
```

Your way is correct but imo using sequences is more "Yamly" :)

(Use the tab correctly in your local editor as I think GH messes with it in reviews)

Same in all the other arrays.

h1dden-da3m0n: while I mostly agree, I would argue for single element arrays I actually prefer the Json array syntax (but to each their own :wink: )

daullmer: I don't have a specific preference, but I think we should stay consistent. Since Vue uses the one @ferferga suggestes (https://github.com/jellyfin/jellyfin-vue/blob/master/.github/workflows/docker.yml) we should go with that

ferferga: I used that syntax in Vue basically because imo it's easier to read short lines and keeps all the file within a column that doesn't exceeds too much in size anywhere.

Otherwise it feels descompensated with the job definition :). But this is super extra nitpicky, no need to go for a "standard" anywhere really, so it's up to you.

daullmer:
Sample run with #65 merged: https://github.com/Ullmie02/jellyfin-ffmpeg/actions/runs/736216721

joshuaboniface (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Push files to repo
+        uses: appleboy/scp-action@v0.1.1
+        with:
+          host: ${{ secrets.REPO_HOST }}
+          username: ${{ secrets.REPO_USER }}
+          password: ${{ secrets.REPO_PASSWORD }}
+          port: 22
+          source: "artifacts/*"
+          target: "/srv/repository/incoming/ffmpeg/${{ github.event.release.tag_name }}"
+          strip_components: 2
```
h1dden-da3m0n: just curious, and not blocking but why scp instead of rsync (then again I only used rsync in the [`android` PR](https://github.com/h1dden-da3m0n/jellyfin-android/blob/1d2d55919a6593dc45229fdafca02aeef2fe7bcc/.github/workflows/app-publish.yaml#L51) due to a comment by ferferga on matrix)

h1dden-da3m0n (commented):
Mind me also asking why you split off the release workflow off the main? (theoretically you could extent the mains triggers and run the `publish` job only if its the correct trigger and with `needs: build`?)

.github/workflows/release.yml
daullmer: First time building a GH actions pipeline from scratch so combining them might be a good idea. I didn't think about that.
I used scp because i was googeling for a task like [this one](https://github.com/jellyfin/jellyfin/blob/master/.ci/azure-pipelines-package.yml#L66) and this task was the one that came up with the most stars. Is there an advantage with using rsync?

joshuaboniface: I've always used `scp` personally. A bit simpler than rsync and we can't always guarantee `rsync` is there (if it is great).

h1dden-da3m0n: there probably is, I haven't played to much with it sadly yet, nor have I looked to deep into it. The sole reason I went for it was ferferga using it in a link he provided on matrix of his server CI revamp. (I'll look for the ref, but IMHO if it works :man_shrugging:)

.github/workflows/main.yml
```diff
@@ -0,0 +1,90 @@
...
+  release: 
+    types:
+      - released
+
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: Build ${{ matrix.release }} ${{ matrix.arch }}
+    runs-on: ubuntu-latest
```
h1dden-da3m0n, line 17:
```suggestion
    runs-on: ubuntu-20.04
```
I have no particular opinion toward using Latest or a fixed version but it would prevent GH from 'complaining' that latest will soon switch to 20.04 and you would be using more up to date tools in the background

.github/workflows/main.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: ./build ${{ matrix.release }} ${{ matrix.arch }} dist
+
+      - name: Upload Packages
+        uses: actions/upload-artifact@v2.2.3
+        with:
+          name: ${{ matrix.release }} ${{ matrix.arch }} package
+          path: dist
+
+  publish:
+    if: github.event_name == 'release' && github.event.action == 'released'
```
h1dden-da3m0n:
```suggestion
    if: ${{ github.event_name == 'release' && github.event.action == 'released' }}
```
again no strong opinion on this either

.github/workflows/release.yml
h1dden-da3m0n: > I've always used `scp` personally. A bit simpler than rsync and we can't always guarantee `rsync` is there (if it is great).

In GH Actions on the [Ubuntu image](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md) it is always there afaik

.github/workflows/main.yml
ferferga: This is about to change this summer so I don't think it's as worth it. Being sticky with latest also reduces future maintenance burden.

h1dden-da3m0n: I know it will change 'soon' :tm:, but it is set to that (not by me) in the CI I have lately worked on (namely android and kotlin-sdk).
However, I am fine either way, just wanted to mention it :wink:

.github/workflows/release.yml
daullmer: @joshuaboniface I can change the workflow to use rsync like @ferferga [here](https://github.com/ferferga/jellyfin/blob/e06cdde3600cbf17a72fc70252067271fb74f82d/.github/workflows/build-packages.yml#L58). However, this action only works with a ssh private key instead of a password.
I have no preference for scp or rsync so I'll let you decide

ferferga (approved):

h1dden-da3m0n (approved):

h1dden-da3m0n (approved):
thanks for catching the rsync arguments :sweat_smile: :+1:

.github/workflows/main.yml
```diff
@@ -0,0 +1,86 @@
...
+          - armhf
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install make and mmv
+        run: sudo apt-get install make mmv
+
+      - name: Install Docker
+        uses: docker-practice/actions-setup-docker@v1
```
h1dden-da3m0n: interesting that this is needed as the virtual environment for the actions has docker, as well as podman, installed.

https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#tools

Note: this is not blocking

h1dden-da3m0n (approved):
wasn't there also one Distro version that would not build due to outdated deps? if so should we not remove them (can be done in another PR)

.github/workflows/main.yml
daullmer: Good catch! It was a long time since I looked at this. I'll see if that can be removed. 

The failing build was for Debian stretch. I wanted to migrate it exactly like it was so I haven't removed it (yet).

h1dden-da3m0n: sounds like a plan, then adding bullseye should happen too in that followup :+1:

daullmer:
I removed the release part of the workflow and will add it again in a later PR because the `collect-ffmpeg.sh` Script is still missing

h1dden-da3m0n (approved):

joshuaboniface (approved):

PR #71, opened 2021-09-06
Drop support for Debian stretch

daullmer:
Drop support for Debian Stretch because a dependency requires nasm 2.13.02 or later. Debian Stretch only has support for 2.12.01.

```
../meson.build:407:16: ERROR: Problem encountered: nasm 2.13.02 or later is required, found nasm 2.12.01
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,7 +22,6 @@ jobs:
         release:
           - bullseye
           - buster
-          - stretch
           - groovy
           - focal
           - bionic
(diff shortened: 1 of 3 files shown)


Bond-009 (approved):

Shadowghost:
https://github.com/jellyfin/jellyfin-ffmpeg/pull/72 should restore Stretch support.

joshuaboniface:
If #72 does indeed fix it, I think we're good to leave it until at least Stretch EOL.

PR #77, opened 2021-11-16
Remove OMX and MMAL from ARM builds

### LINKED ISSUE #32: jellyfin-ffmpeg for armhf assumes Raspberry Pi device
Hi, trying to run jellyfin on an Odroid XU4 running Ubuntu 18.04.4 LTS. When installed, jellyfin pulls in jellyfin-ffmpeg and installs it. Attempting to play back video that requires transcoding results in many errors such as:
```
[2020-03-22 07:14:17.676 +00:00] [ERR] Error in "ffprobe"
System.Text.Json.JsonException: The input does not contain any JSON tokens. Expected the input to start with a valid JSON token, when isFinalBlock is true. Path: $ | LineNumber: 0 | BytePositionInLine: 0.
```
When I attempt to run jellyfin-ffmpeg manually, it dies due to a missing dependency:
```
root@odroid:/var/log/jellyfin# /usr/lib/jellyfin-ffmpeg/ffmpeg
/usr/lib/jellyfin-ffmpeg/ffmpeg: error while loading shared libraries: libmmal_core.so: cannot open shared object file: No such file or directory
``` 

As far as I can tell, libmmal_core.so only exists on the Raspberry Pi platform. Why is the armhf build of jellyfin-ffmpeg depending upon a raspberry pi only library?

ferferga:
For 10.8, I think we should remove all of the Pi-specific hardware acceleration stuff: it never worked properly (besides a very specific configuration set) and it's better to integrate something that truly works fine instead of the patchwork that current MMAL and OMX is.

Raspberry Pi development is transitioning towards V4L2 for hardware decoding/encoding and MMAL/OMX are in a deprecation path. They also only worked properly in Raspbian Stretch and Pi3 devices, other configurations produced mixed results. Right now, in Bullseye, ffmpeg and ffprobe doesn't start at all. 

With this patch, Pis will do software encoding/decoding (which is what most people ended using up anyway) until there's a better hardware acceleration path in the future.

Fixes jellyfin/jellyfin#32 jellyfin/jellyfin#75

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,8 +1,14 @@
-jellyfin-ffmpeg (4.4.1-2) unstable; urgency=medium
+jellyfin-ffmpeg (4.4.1-3) unstable; urgency=medium
 
   * Remove OMX and MMAL from ARM builds
 
  -- ferferga <ferferga@hotmail.com>  Sun, 26 Dec 2021 19:58:10 +0100
+
+jellyfin-ffmpeg (4.4.1-2) unstable; urgency=medium
+
+  * Enable Link Time Optimization (LTO)
+
+ -- Bond-009 <bond.009@outlook.com>  Thu Apr 8 13:57:13 2021 +0200
  
 jellyfin-ffmpeg (4.4.1-1) unstable; urgency=medium
 
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,9 @@
+jellyfin-ffmpeg (4.4.1-2) unstable; urgency=medium
+
+  * Remove OMX and MMAL from ARM builds
+
+ -- ferferga <ferferga@hotmail.com>  Sun, 26 Dec 2021 19:58:10 +0100
+ 
 jellyfin-ffmpeg (4.4.1-1) unstable; urgency=medium
 
   * New upstream version 4.4.1
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -19,10 +19,12 @@ ENV CFLAGS="-I${TARGET_DIR}/include $CFLAGS"
 
 # Prepare Debian build environment
 RUN apt-get update \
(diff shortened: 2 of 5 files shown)


nyanmisaka:
+1 for this. Link to the issue https://github.com/jellyfin/jellyfin-ffmpeg/issues/310

ferferga:
jellyfin/jellyfin#29 has useful resources as well

crobibero (approved):

nyanmisaka (approved):
LTGM

```
error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.
fatal: the remote end hung up unexpectedly
fatal: early EOF
```

debian/changelog
```diff
@@ -1,3 +1,8 @@
+jellyfin-ffmpeg (4.4.1-2) unstable; urgency=medium
+
+  * Remove OMX and MMAL from ARM builds
+
+ -- ferferga <ferferga@hotmail.com>  Sun, 26 Dec 2021 19:58:10 +0100
```
joshuaboniface, line 5: One extra newline here or it will error.

joshuaboniface (commented):
LGTM aside from the mentioned comment.

nyanmisaka:
The github server seems to be unstable these days.
Is there a better way to make git retry after losing the connection?

ferferga:
@nyanmisaka Thats what I'm trying to fix with the latest commits, but I think it's a bug in the git installed, not 100% related to GitHub (which also plays a role as some of these repos clone rather slow).

ferferga:
@nyanmisaka Checks are passing now, stretch is failing for a completely different reason. Probably jellyfin/jellyfin#65 ? [First checks with fails in that strecth suite comes from that merge](https://github.com/jellyfin/jellyfin-ffmpeg/runs/4639723973?check_suite_focus=true)

nyanmisaka:
LTO is not supported in old GCC. We should retire stretch.

Shadowghost:
I'm all in for deprecating Debian stretch. We should only support Debian stable and oldstable imo (same for the other packages).

PR #80, opened 2022-01-02
Remove Debian Stretch

crobibero:
Removes "oldoldstable"

### DIFF
--- a/Dockerfile.make
+++ b/Dockerfile.make
@@ -1,6 +1,6 @@
 #!/usr/bin/make
-DISTRO=stretch
-GCC_VER=6
+DISTRO=bullseye
+GCC_VER=10
 ARCH=amd64
 .PHONY: Dockerfile
 Dockerfile: Dockerfile.in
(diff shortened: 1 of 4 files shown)


nyanmisaka (approved):

Dockerfile.make
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/make
-DISTRO=stretch
-GCC_VER=6
+DISTRO=bullseye
```
ferferga, line 2: *Quick pass in phone, so not fully aware of all the casuistics*

I think changing this doesn't matter? Below the dockerfiles used for compilation are established the correct target and GCC version. Why not use those there as well? 
Any idea why this file exists in the first place?
If this is needed, maybe setting it to a default value we don't have to constantly update (like stable, so we don't tie it up to any specific release) it's a better idea?

joshuaboniface: It's just a default that's set in case the `make` command is run without args (`make -f Dockerfile.make DISTRO=${release} GCC_VER=${gcc_version} ARCH=${arch}` in `build`); changing them shouldn't have any negative effects.

joshuaboniface (approved):

PR #93, opened 2022-02-25
Add libva-utils (vainfo)

Shadowghost:
Since the distros `vainfo` package might be incompatible with the libva version we use, we're including it into our debian package.

Changes:
* Add building of vainfo (via libva-utils) to the build script (amd64 only)
* Add required licenses to the package config

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -115,6 +115,17 @@ prepare_extra_amd64() {
     popd
     popd
 
+    # Download and install libva-utils
+    pushd ${SOURCE_DIR}
+    git clone --depth=1 https://github.com/intel/libva-utils
+    pushd libva-utils
+    ./autogen.sh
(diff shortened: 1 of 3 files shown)


docker-build.sh
```diff
@@ -179,7 +178,19 @@ prepare_extra_amd64() {
     popd
     popd
     popd
+
+    # Download and install libva-utils
```
nyanmisaka: LGTM but can you move this field to right after the `# Download and install libva`? The relationship between their version numbers seems to be very close.

nyanmisaka (approved):

docker-build.sh
Shadowghost: Done. I also removed the explicit version of libva-utils since we use master of libva too.

joshuaboniface (approved):

PR #96, opened 2022-02-28
Rework CI

Shadowghost:
Goals of this PR:
* automatically build all packages with a unified meta workflow
* append all artifacts to a github release
* automatically upload releases to the repo server and sync the mirrors


.github/workflows/build.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - jellyfin
+    paths-ignore:
+      - '**/*.md'
+
+jobs:
+  build_debian:
+    uses: jellyfin/jellyfin-ffmpeg/.github/workflows/_meta.yaml@master
+    with:
+      distro: 'debian'
+      codenames: "[\"buster\", \"bullseye\"]"
```
h1dden-da3m0n: could you not use `'["buster", "bullseye"]'` instead of the escape? would look cleaner IMHO if possible

.github/workflows/_meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Build Linux
+        if: ${{ contains(inputs.distro, 'debian') || contains(inputs.distro, 'ubuntu') }}
+        run: ./build ${{ matrix.release }} ${{ matrix.arch }} dist
+
+      - name: Build Windows
+        if: ${{ contains(inputs.distro, 'windows') }}
+        run: ./build-win64 dist
+
+      - name: Upload Linux Artifacts
+        uses: actions/upload-artifact@v2
```
h1dden-da3m0n: you pinned the versions for the other actions but not these and checkout?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,78 @@
+name: '🏗️ Build and Publish jellyfin-ffmpeg'
```
h1dden-da3m0n:
```suggestion
name: '🚀 Build and Publish jellyfin-ffmpeg'
```

Additionally, any particular reason you explicitly exclude pre-releases?

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,78 @@
...
+        uses: appleboy/ssh-action@v0.1.4
+        env:
+          DISTRO: ${{ matrix.distro }}
+          RELEASE: ${{ github.release.tag_name }}
+        with:
+          host: ${{ secrets.deploy-host }}
+          username: ${{ secrets.deploy-user }}
+          key: ${{ secrets.deploy-key }}
+          script_stop: true
+          envs: DISTRO, RELEASE
```
h1dden-da3m0n: this certainly works but you could also simply use the GH expression in the script section as it will get evaluated and interpolated before it gets send of to be run. (that is also applicable for the regular `run` blocks)

.github/workflows/_meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+        with:
+          upload_url: ${{ github.event.release.upload_url }}
+          asset_path: ./*
+
+      - name: Upload Assets
+        uses: burnett01/rsync-deployments@5.2
+        with:
+          switches: -vrptz
+          path: ./*
+          remote_path: /srv/repository/releases/server/${inputs.distro}/versions/jellyfin-ffmpeg/${github.release.tag_name}/
```
h1dden-da3m0n: I am not 100% sure, but think the repo dir should exist already, but if not maybe run an ssh script b4 this to ensure the folder exists.

h1dden-da3m0n (commented):
LGTM overall, but a few comments

.github/workflows/publish.yaml
Shadowghost: Reason is that the prerelease would be pushed to the repo server - and we don't want a pre-release on out stable repos.

.github/workflows/publish.yaml
Shadowghost: I just adapted what you did here: https://github.com/jellyfin/jellyfin-meta-plugins/blob/master/.github/workflows/publish.yaml

.github/workflows/_meta.yaml
Shadowghost: Yeah that was something I thought of too. I guess it's a good idea to do it nonetheless.

.github/workflows/build.yaml
Shadowghost: I took this from here where it was explictily escaped:
https://github.community/t/reusable-workflow-with-strategy-matrix/205676

but I guess you can do it properly without the escape. I'll test it.

.github/workflows/_meta.yaml
Shadowghost: Pinned those too.

.github/workflows/publish.yaml
h1dden-da3m0n: makes sense, so this is kind of a 'preparation' for if we ever actively intent to create per-releases?

.github/workflows/publish.yaml
h1dden-da3m0n: yeah, I did that there to have the 2 variables out of the script and within the Workflow. I however also have expression syntax in there.
Anyway, this is just a comment as to what is possible, use what you find convenient and passing the envs over might be the easier to write/read way anyway.

Shadowghost: I changed it nonetheless. We can change it again if it does not work

.github/workflows/publish.yaml
Shadowghost: [source](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#release) I changed the trigger to `released` that should exclude all the cases we don't want.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,71 @@
...
+        uses: appleboy/ssh-action@v0.1.4
+        with:
+          host: ${{ secrets.deploy-host }}
+          username: ${{ secrets.deploy-user }}
+          key: ${{ secrets.deploy-key }}
+          script_stop: true
+          script: |-
+            if [ "${matrix.distro}" != "windows" ]; then
+              sudo find /srv/repository/releases/server/${matrix.distro}/ -type l -name "jellyfin-ffmpeg_*" -exec rm {} \;
+              reprepro -b /srv/repository/${matrix.distro} deleteunreferenced
```
joshuaboniface, line 70: These two lines (deleteunreferenced and export) have to be after the includedeb.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,71 @@
...
+              reprepro -b /srv/repository/${matrix.distro} export
+              find "/srv/repository/${matrix.distro}" -name "*.deb" | while read file; do
```
joshuaboniface: This should be a similar consruct to the find above, to ensure it's looking in the right place, or even more specific right down into the exact folder where the upload puts the files.

I'd actually much prefer this upload into a temporary location (i.e. `/srv/repository/incoming/ffmpeg`) then import those into `reprepro`, then finally move them to the final place and update symlinks. It seems safer that way.

joshuaboniface (changes requested):

.github/workflows/publish.yaml
h1dden-da3m0n: and while we here `${{matrix.distro}}` if you want to use the expression syntax, otherwise this would be plain bash syntax and break when it reaches this point.

`${{ <expression> }}` GitHub Action Expression
`${<param>}` Bash Parameter expression

.github/workflows/publish.yaml
Shadowghost: I think I got it right this time. Sorry for the force push, I fucked up some things locally...

joshuaboniface (approved):
Looks good :+1:

PR #101, opened 2022-03-02
Allow configuration of used driver via ENV variable

Shadowghost:
Currently our shipped `vainfo` or any other program using our ffmpeg can't explicitly set the used driver via the ENV var. I don't see a reason why we should explicitly prevent this, so I added it back.

To explicitly set this environment for Jellyfin only (e.g. multiple GPUs, one for transcoding one for something else) the variable was renamed from `LIBVA_DRIVER_NAME` to `LIBVA_DRIVER_NAME_JELLYFIN`.

### DIFF
--- a/docker-build.sh
+++ b/docker-build.sh
@@ -106,7 +106,7 @@ prepare_extra_amd64() {
     git clone --depth=1 https://github.com/intel/libva
     pushd libva
     sed -i 's|getenv("LIBVA_DRIVERS_PATH")|"/usr/lib/jellyfin-ffmpeg/lib/dri:/usr/lib/x86_64-linux-gnu/dri:/usr/lib/dri:/usr/local/lib/dri"|g' va/va.c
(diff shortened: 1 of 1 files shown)


nyanmisaka:
On a Nvidia prime laptop with both [`nvidia_dri_video`](https://github.com/elFarto/nvidia-vaapi-driver) and `iHD_dri_video` driver installed, if user set `LIBVA_DRIVER_NAME=nvidia` explicitly to global system to enable hw decoding in firefox/chrome, then the jellyfin-ffmpeg cannot initialize iHD driver for the integrated iGPU/APU to transcoding.

You can rename it to something like `LIBVA_DRIVER_NAME_JELLYFIN`.

Shadowghost:
Sounds like a reasonable approach for this kind of situations to me. I changed the code in regards to this.

crobibero (approved):

PR #107, opened 2022-03-06
Fix bash shenanigans in CI

Shadowghost:


.github/workflows/_meta.yaml
```diff
@@ -96,9 +96,11 @@ jobs:
 
       - name: Prepare Release Assets
         run: |-
-          find ./ -type f \( -name "*.deb" -o -name "*.zip" \) | while read file; do
-            sha256sum ${file#./} >> ${file%.*}.sha256
+          cd artifact
+          find . -type f \( -name "*.deb" -o -name "*.zip" \) | while read file; do
```
joshuaboniface: This will still have the same problem since `find` will append `./` to the start.

Should just be `for file in *.deb *.zip; do`.

joshuaboniface (changes requested):

.github/workflows/_meta.yaml
joshuaboniface: Also should likely use `pushd`/`popd` instead of `cd`, but that's down to preference and shouldn't matter in this case.

joshuaboniface (approved):

PR #109, opened 2022-03-06
Another try to fix CI

Shadowghost:
* properly generate checksums only for existing files
* only upload relevant files to release

### DIFF
--- a/.github/workflows/_meta.yaml
+++ b/.github/workflows/_meta.yaml
@@ -92,7 +92,7 @@ jobs:
       - name: Prepare Release Assets
         run: |-
           pushd artifact
-          for file in *.deb *.zip; do
+          find * -type f \( -name "*.deb" -o -name "*.zip" \) | while read file; do
             sha256sum "${file}" | tee "${file}.sha256sum"
           done
           popd
@@ -101,7 +101,10 @@ jobs:
         uses: shogo82148/actions-upload-release-asset@v1.5.0
         with:
           upload_url: ${{ github.event.release.upload_url }}
-          asset_path: ./artifact/*
+          asset_path: |
+            ./artifact/*.zip'
+            ./artifact/*.deb'
+            ./artifact/*.sha256sum'
 
       - name: Make Sure Release Directory Exists
         uses: appleboy/ssh-action@v0.1.4


crobibero (approved):

.github/workflows/_meta.yaml
```diff
@@ -92,7 +92,7 @@ jobs:
       - name: Prepare Release Assets
         run: |-
           pushd artifact
-          for file in *.deb *.zip; do
+          find * -type f \( -name "*.deb" -o -name "*.zip" \) | while read file; do
```
oddstr13, line 95: While I would hate whoever made uppercase bits in these file extensions, it may be worth using `-iname` for case insensitive filename matching.

There is also a difference between `find *` and `find .`, dotfiles are not matched by `*`.
Not that we'd ever want dotfiles as actual artifacts tho, just pointing it out.

```suggestion
          find * -type f \( -iname "*.deb" -or -iname "*.zip" \) | while read file; do
```
I prefer the long version of or, as it's more clear on what it does without being familiar with find (but it isn't POSIX, which shouldn't matter as we're running this under Ubuntu).

.github/workflows/_meta.yaml
```diff
@@ -101,7 +101,10 @@ jobs:
         uses: shogo82148/actions-upload-release-asset@v1.5.0
         with:
           upload_url: ${{ github.event.release.upload_url }}
-          asset_path: ./artifact/*
+          asset_path: |
+            ./artifact/*.zip'
+            ./artifact/*.deb'
+            ./artifact/*.sha256sum'
```
oddstr13, line 107: Copypasta?
```suggestion
            ./artifact/*.zip
            ./artifact/*.deb
            ./artifact/*.sha256sum
```

Depending on how the glob works here, it may not pick up files in sub-directories.
I'm not sure if that ever matters here.

oddstr13 (commented):
`asset_path` needs fixing.
The `find` suggestions are just that, suggestions.

.github/workflows/_meta.yaml
Shadowghost: I agree on using `iname` but the use of `*` is intentional to strip the leading `./` from the result of find.

.github/workflows/_meta.yaml
Shadowghost: Bad copypasta and originally it was intentional to only grab on one level but I guess it doesn't hurt picking up recursivly.

Fixes didn't work anyway so I'll change it in the next try...

PR #120, opened 2022-03-23
Bump dependencies and version

Shadowghost:
* Bump dependencies
* Add changelog
* Bump ffmpeg version to -4

Note:
* Do not bump gmmlib because of https://github.com/intel/gmmlib/issues/95 (it's fixed on gmmlib master but affects anything after 22.0.2)

### DIFF
--- a/build-win64
+++ b/build-win64
@@ -10,7 +10,7 @@ done
 
 # Use the latest distro for toolchains
 distro="ubuntu:jammy"
-ffrevison="3"
+ffrevison="4"
 image_name="jellyfin-ffmpeg-build-windows-win64"
 package_temporary_dir="$( mktemp -d )"
 current_user="$( whoami )"
--- a/build.yaml
+++ b/build.yaml
@@ -1,7 +1,7 @@
 ---
 # We just wrap `build` so this is really it
 name: "jellyfin-ffmpeg"
-version: "4.4.1-3"
+version: "4.4.1-4"
 packages:
   - buster-amd64
   - buster-armhf
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+jellyfin-ffmpeg (4.4.1-4) unstable; urgency=medium
+
+  * Update dependencies (#120)
+  * Update dav1d to 1.0.0 (#118)
+  * fix libass AR and blur issue (#119)
+
+ -- Joshua Boniface <joshua@boniface.me>  Wed, 23 Mar 2022 01:02:21 -0500
+ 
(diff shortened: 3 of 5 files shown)


nyanmisaka:
Have you verified gmmlib 22.1.1 with compute-runtime's gmmlib 22.0.2?

The last time I tried 22.0.3 it broke the command `ffmpeg -v verbose -init_hw_device vaapi=va:,driver=iHD -init_hw_device qsv=qs@va -init_hw_device opencl=ocl@va` and throw segfault error.

nyanmisaka:
Keeping the version of `nv-codec-headers` at `11.0.10.1` is to provide forward compatibility where possible.
`11.0.10.1` requires 455/456 driver but the latest `11.1.5.1` requires 470/471 driver. Both of them support AV1 decoding.

https://github.com/FFmpeg/nv-codec-headers/blob/n11.0.10.1/README
https://github.com/FFmpeg/nv-codec-headers/blob/n11.1.5.1/README

Shadowghost:
> Have you verified gmmlib 22.1.1 with compute-runtime's gmmlib 22.0.2?
> 
> The last time I tried 22.0.3 it broke the command `ffmpeg -v verbose -init_hw_device vaapi=va:,driver=iHD -init_hw_device qsv=qs@va -init_hw_device opencl=ocl@va` and throw segfault error.

It segfaulted for me too with 22.1.1 (there is a commit to fix the segfault in master, but not in a release version).

I'm currently trying 22.0.3. If that segfaults too I'll just let it be at 22.0.2 (though I don't see why 22.0.3 would break compatibility with 22.0.2 since ABI changes require a major version bump as per readme)

nyanmisaka:
> it's related to GmmLib ABI broken by https://github.com/intel/gmmlib/commit/8dcebfefaaacff0cbc262fe050b2b0fd060898d0, please update media code after https://github.com/intel/media-driver/commit/79c20b78a and gmmlib code after https://github.com/intel/gmmlib/commit/29152fc6d39f87983b176664f225d42af039f716

The regression code was added in gmmlib 22.0.3. https://github.com/intel/media-driver/issues/1364#issuecomment-1075833402
I have reverted it in https://github.com/jellyfin/jellyfin-ffmpeg/pull/112

Shadowghost:
I reverted to 22.0.2 but I'll check again after they released the new tag.
Should be fine to use the tag when available. I built it with gmmlib master and everything seems to work fine (105 fps on a 4k HDR SRT -> 1080p SDR stream on an UHD630)

joshuaboniface (approved):

PR #143, opened 2022-05-06
Add generic windows ffmpeg symlink to CI

Shadowghost:


joshuaboniface (commented):
Looks good. Normally I'd recommend against symlinking to a glob but this *should* be good every time.

joshuaboniface (approved):

crobibero (approved):

PR #156, opened 2022-06-12
Fix some issues and update dependencies

nyanmisaka:
**Changes**
- Fix the unsupported subscripted access on nVidia (we don't use opencl on nVidia though)
- Fix using hwupload with hwmap on d3d11va
- Update dependencies
- Bump version to 5.0.1-6
- Add params to trace the symlink err on publishing
- Force reporting packed headers are supported on AMD VAAPI (it's a hack to workaround something like https://github.com/jellyfin/jellyfin/issues/7642)

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -71,6 +71,8 @@ jobs:
           key: ${{ secrets.DEPLOY_KEY }}
           script_stop: true
           script: |-
+            set -o errexit
+            set -o xtrace
             tag="${{ github.event.release.tag_name }}"
             version="${tag#v}"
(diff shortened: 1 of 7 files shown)


nyanmisaka:
```
fatal: unable to access 'https://gitlab.freedesktop.org/wtaymans/fdk-aac-stripped.git/': The requested URL returned error: 504
```
The freedesktop repo seems to be undergoing maintenance or it's just down.

Shadowghost (approved):

nyanmisaka:
I've changed to get `libdrm` and `mesa` from source tar but it seems that the latest `fdk-aac-stripped` is only available on gitlab.freedesktop.org. Now it's fully recovered.

PR #161, opened 2022-06-22
Pin the zimg version

nyanmisaka:
**Changes**
- Pin the zimg version

It seems that their master branch is broken.


tandy-1000:
Ah, nicely spotted!

Shadowghost (approved):

PR #187, opened 2022-09-30
New upstream version 5.1.2 and enable SVT-AV1 encoder

nyanmisaka:
**Changes**
- New upstream version 5.1.2
- Enable [SVT-AV1](https://gitlab.com/AOMediaCodec/SVT-AV1/-/blob/master/Docs/Ffmpeg.md) encoder for amd64
- Remove unneed assertion from AMF encoder
- Disable reshape if DOVI require an EL
- Update dependencies
- Bump version to 5.1.2-1


Shadowghost (approved):

debian/control
```diff
@@ -5,7 +5,7 @@ Maintainer: Jellyfin Packaging Team <packaging@jellyfin.org>
 Uploaders: Jellyfin Packaging Team <packaging@jellyfin.org>
 Rules-Requires-Root: no
 Homepage: https://ffmpeg.org/
-Standards-Version: 5.1.1
```
Shadowghost, line 8: Is this intentional?

nyanmisaka: Yes. `Standards-Version` should be these: https://www.debian.org/doc/debian-policy/upgrading-checklist.html

nyanmisaka:
```
The following packages have unmet dependencies:
 linux-libc-dev : Breaks: linux-libc-dev:armhf (!= 4.19.260-1) but 4.19.249-2 is to be installed
 linux-libc-dev:armhf : Breaks: linux-libc-dev (!= 4.19.249-2) but 4.19.260-1 is to be installed
```

~~Dependencies issue from upstream again......~~

debian/changelog
```diff
@@ -1,3 +1,12 @@
+jellyfin-ffmpeg (5.1.2-1) unstable; urgency=medium
+
+  * New upstream version 5.1.2
+  * Enable SVT-AV1 encoder for amd64
+  * Remove unneeded asseetion from AMF encoder
```
Shadowghost:
```suggestion
  * Remove unneeded assertion from AMF encoder
```

Bond-009 (approved):

PR #197, opened 2022-10-29
Hotfix for Mesa driver build

nyanmisaka:
**Changes**
- Hotfix for Mesa driver build


nyanmisaka:
We may need to rebuild the docker image for 10.8.6.

nielsvanvelzen:
10.8.7 will be released soon so I don't think rebuilding the 10.8.6 container is necessary.

Shadowghost (approved):

PR #223, opened 2023-03-02
New upstream version 6.0

nyanmisaka:
Deb package is renamed to `jellyfin-ffmpeg6`.

**Changes**
- New upstream [version 6.0](http://ffmpeg.org/download.html#releases)
- Update patches for 6.0
- Bump version to 6.0-1
- Drop the EOL Ubuntu bionic
- Enable Debian bookworm and Ubuntu lunar

**Important**
- `libmfx` is superseded by `libvpl` for enabling the QSV AV1 encoder. This change is seamless.
- [nv-codec-headers](https://github.com/FFmpeg/nv-codec-headers) is updated to meet the requirement of NVENC AV1 encoder,
 the minimum required NVIDIA driver versions have been changed to 520/522 series.

```
Corresponds to Video Codec SDK version 12.0.16.

Minimum required driver versions:
Linux: 520.56.06 or newer
Windows: 522.25 or newer
```


Shadowghost (approved):
We should not forget to update the package name in metapackages and the dockerfile in the server repo.

build.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 # We just wrap `build` so this is really it
 name: "jellyfin-ffmpeg"
-version: "5.1.2-8"
+version: "6.0-1"
```
Shadowghost, line 4:
```suggestion
version: "6.0.0-1"
```

We should stay consistent in the versioning.

nyanmisaka: How this affects our workflow? I thought 6.0 and 6.0.0 are equal.

If this is necessary, does it mean we have to modify the package version to 6.0.0-1?

Shadowghost: I'm not sure if Debian likes it if you swith between `Major.Minor` and `Major.Minor.Patch`

IMHO we should stay consistent and use `Major.Minor.Patch` here and in the package version.

Shadowghost: For completeness: Debian has no problem with this, so ignore the comment.

PR #247, opened 2023-05-24
Enable ffplay generation in the ffmpeg build

satmandu:
**Changes**
Adds back the `ffplay` generation in the `ffmpeg` build, which was previously disabled. This allows comparison testing of the `jellyfin-ffplay` binary with system `ffplay` builds.

### DIFF
--- a/debian/jellyfin-ffmpeg6.install
+++ b/debian/jellyfin-ffmpeg6.install
@@ -1,4 +1,5 @@
 ffmpeg usr/lib/jellyfin-ffmpeg
+ffplay usr/lib/jellyfin-ffplay
 ffprobe usr/lib/jellyfin-ffmpeg
 libavcodec/libavcodec.so* usr/lib/jellyfin-ffmpeg/lib
 libavdevice/libavdevice.so* usr/lib/jellyfin-ffmpeg/lib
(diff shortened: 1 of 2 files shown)


nyanmisaka:
The patches we added do nothing for ffplay, and the Jellyfin server won't use it.
So enabling it will only increase the package size further.
It is recommended to use ffplay that comes with distro, there is no difference between the two.

satmandu:
Some distributions (e.g. RaspiOS ) carry their own ffmpeg patches, so it would be useful to compare their ffplay with the the ffplay from jellyfin-ffmpeg. (Looking at frame rates and so on, especially when trying different acceleration options.)

It won't bloat the package too much though. On my arm64 Ubuntu system:

```
satadru@rpi4b ~$ ls -aFlh `which ffplay`
-rwxr-xr-x 1 root root 195K Feb 17 17:28 /usr/bin/ffplay*
```

I fully grant though that ffplay is not used at all in Jellyfin otherwise...

nyanmisaka:
Perhaps it would be better to fork this repo and use Github Action to build your custom version online if it's just for testing and experimentation? to avoid having everyone install extra copies of ffplay. We also have portable versions of jellyfin-ffmpeg which use static linking so ffplay can be very large ~70MB.

satmandu:
Thanks. I'll do that!

satmandu:
(And it looks like the output could be quite a bit larger as `ffplay` requires libsdl2-dev to be pulled in for the build.)

PR #277, opened 2023-08-30
[5.1] Add icon for Windows version FFmpeg

nyanmisaka:
**Changes**
- Add icon for Windows version FFmpeg

ff

Cosmetic changes to make it easier for users to distinguish from vanilla FFmpeg builds.
The icon color comes from https://jellyfin.org/docs/general/contributing/branding


Shadowghost (approved):
We should add it to the UX repo too..

Might also inlcude it into the readme (we still need to overault it anyway)

nyanmisaka:
> We should add it to the UX repo too..

Done in https://github.com/jellyfin/jellyfin-ux/pull/56

PR #287, opened 2023-10-09
Add Ubuntu Mantic to build system

oddstr13:
**Changes**

Add Ubuntu Mantic to build system.

It *does* look like this might be all that's required to actually create a repo and publish to it, but I am not 100% certain.

**Issues**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -30,7 +30,7 @@ jobs:
     uses: ./.github/workflows/_meta.yaml
     with:
       distro: 'ubuntu'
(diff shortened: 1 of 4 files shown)


nyanmisaka:
https://github.com/jellyfin/jellyfin-ffmpeg/blob/3c25eff1550cfc5cfa333ef810517281277a56a1/.github/workflows/publish.yaml#L90

I already have this in my [PR](https://github.com/jellyfin/jellyfin-ffmpeg/pull/283/commits/c01c2f41a702874077333514a2fbe3306638a759). The problem is that the codename is missing from the server so that it is impossible to upload the github build to the server and update the apt repo by using `reprepro`.

PR #337, opened 2024-02-12
Refactor Dockerfile

Gauvino:
Refactor Dockerfiles for better syntax


Gauvino:
Remake the PR cause they're is to many commit on the early one and abandonned the idea to make the env proper cause it's always failing

nyanmisaka:
IIRC removing `--no-install-recommends` should fix the problem.

These images are not intended to be used as final products, so size and cleanup should not be too much of a concern.

Gauvino:
Everything should now be good and built nicely and lovely

nyanmisaka (approved):

PR #345, opened 2024-03-04
feat: add macOS builder script for CI

### LINKED ISSUE #339: Road to macOS builds

gnattu:
**Changes**

Preliminary scripts for macOS builder. There are enough difference that made me to make a dedicated script.

One thing to note is that the arm64 host is macOS 14 exclusive, which means the binary built for arm64 will use newer toolchain than the x86 one using macOS 12 toolchain.

The script currently is only good for CI clean build, but documented the caveats in case someone want to build locally.

**Issues**

Fixes #339


builder/scripts.d/20-libiconv.sh
```diff
@@ -5,6 +5,8 @@ SCRIPT_COMMIT="v1.17"
 SCRIPT_TAGFILTER="v?.*"
 
 ffbuild_enabled() {
+    # iconv is macOS built-in
+    [[ $TARGET == mac* ]] && return -1
```
nyanmisaka: I understand what you're doing here, but it's counter-intuitive.
Maybe return 0 directly for `mac*` in `ffbuild_dockerbuild()` and leave `ffbuild_unconfigure()` untouched?

nyanmisaka (approved):
LGTM. Still need to wait for @Shadowghost to take a closer look at the changes in GH actions.

Gauvino (approved):
CI and typos LGTM

.github/workflows/_meta_mac_portable.yaml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Download Artifacts
+        uses: actions/download-artifact@c850b930e6ba138125429b7e5c93fc707a7f8427 # v4.1.4
+        with:
+          name: mac-${{ matrix.arch }}-portable
+          path: artifact
+
+      - name: Prepare Release Assets
+        run: |-
+          pushd artifact
+          find * -type f \( -name "*.xz" -o -name "*.zip" \) | while read file; do
```
Shadowghost:
```suggestion
          find * -type f \( -name "*.xz" -o -name "*.zip" \) | while read file; do
```

Isn't the builder producing tar archives? If it only produces tar and already creates the sha256sum you can remove this step, otherwise adopt to the archive extension - also in the next step.

gnattu: This is copied from our other release pipeline, and the builder produces tar.xz archives. I've seen our other build scripts producing similar checksum files as well, does that mean this step is redundant?

Shadowghost: If we already have sha256sum files it is redundant. But I think we do not create them for windows.

Shadowghost (approved):
Should be good to go, If something in the workflows is broken, we'll see it only on publish anyways.

PR #352, opened 2024-03-09
fix VT tone mapping and add transpose_vt

gnattu:
**Changes**

This PR:
- Added a back-ported `transpose_vt` filter which uses CoreImage that works on macOS 12
- Added format option for `scale_vt` filter. Currently only supports `nv12` and `p010`

**Issues**

### DIFF
--- a/debian/patches/0061-backport-scale-vt.patch
+++ b/debian/patches/0061-backport-scale-vt.patch
@@ -90,7 +90,7 @@ diff --git a/libavfilter/vf_scale_vt.c b/libavfilter/vf_scale_vt.c
 new file mode 100644
 --- /dev/null	(revision c2c96c4c2419859c4d1b11e6f907e58afb6dfa3c)
 +++ b/libavfilter/vf_scale_vt.c	(revision c2c96c4c2419859c4d1b11e6f907e58afb6dfa3c)
-@@ -0,0 +1,292 @@
+@@ -0,0 +1,293 @@
 +/*
 + * Copyright (c) 2023 Zhao Zhili <zhilizhao@tencent.com>
 + *
@@ -141,6 +141,7 @@ new file mode 100644
 +static const enum AVPixelFormat supported_formats[] = {
 +    AV_PIX_FMT_NV12,
 +    AV_PIX_FMT_P010,
++    AV_PIX_FMT_NONE,
 +};
 +
 +static int format_is_supported(enum AVPixelFormat fmt)
@@ -306,7 +307,7 @@ new file mode 100644
 +    hw_frame_ctx_in = (AVHWFramesContext *)inlink->hw_frames_ctx->data;
 +
 +    out_format = (s->format == AV_PIX_FMT_NONE) ? hw_frame_ctx_in->sw_format : s->format;
-+    if (!format_is_supported(out_format)) {
++    if (!format_is_supported(s->format)) {
 +        av_log(s, AV_LOG_ERROR, "Unsupported output format: %s\n",
 +               av_get_pix_fmt_name(out_format));
 +        return AVERROR(ENOSYS);
--- a/debian/patches/0061-backport-scale-vt.patch
+++ b/debian/patches/0061-backport-scale-vt.patch
@@ -294,7 +294,7 @@ new file mode 100644
 +    outlink->hw_frames_ctx = av_hwframe_ctx_alloc(hw_frame_ctx_in->device_ref);
 +    hw_frame_ctx_out = (AVHWFramesContext *)outlink->hw_frames_ctx->data;
 +    hw_frame_ctx_out->format = AV_PIX_FMT_VIDEOTOOLBOX;
-+    hw_frame_ctx_out->sw_format = hw_frame_ctx_in->sw_format;
++    hw_frame_ctx_out->sw_format = s->colour_primaries == AVCOL_PRI_BT709 ? AV_PIX_FMT_NV12 : hw_frame_ctx_in->sw_format;
 +    hw_frame_ctx_out->width = outlink->w;
 +    hw_frame_ctx_out->height = outlink->h;
 +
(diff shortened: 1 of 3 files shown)


debian/patches/0061-backport-scale-vt.patch
```diff
@@ -294,7 +294,7 @@ new file mode 100644
 +    outlink->hw_frames_ctx = av_hwframe_ctx_alloc(hw_frame_ctx_in->device_ref);
 +    hw_frame_ctx_out = (AVHWFramesContext *)outlink->hw_frames_ctx->data;
 +    hw_frame_ctx_out->format = AV_PIX_FMT_VIDEOTOOLBOX;
-+    hw_frame_ctx_out->sw_format = hw_frame_ctx_in->sw_format;
++    hw_frame_ctx_out->sw_format = s->colour_primaries == AVCOL_PRI_BT709 ? AV_PIX_FMT_NV12 : hw_frame_ctx_in->sw_format;
```
nyanmisaka: It is better to expose a format option to make it consistent with other hardware `scale_*` filters. I don't think upstream likes this implicit format conversion either, unless it's documented somewhere that Apple explicitly requires this.

gnattu: In its current form, this is a quick fix for our downstream issue. We don't have many other options, as formats other than NV12 and P010 may make the encoder unhappy and causing frame ordering issue. However, I do agree that an option is needed for an upstream patch.

nyanmisaka: Yeah it's a stopgap. Actually a potential breakage is that `tonemap_opencl` requires `p010le` input. This may fail when used with `scale_vt` downscaling since we can't set any format option.

gnattu: I noticed that, so I added this. When we are not using `scale_vt` to perform tone mapping, the output frame will retain the format as is. Therefore, the OpenCL tone mapping can still receive a P010 input. However, when we are setting color primaries, it means that we are using `scale_vt` itself to perform tone mapping, and its output is unlikely to be used with OpenCL tone mapping.

nyanmisaka: > the output frame will retain the format as is. Therefore, the OpenCL tone mapping can still receive a P010 input

That's because we currently only have hardware decoding enabled for [4:2:0 sub-sampling](https://github.com/jellyfin/jellyfin/blob/d10ad6c383954e6acb76581d5459aa9740c66688/MediaBrowser.Controller/MediaEncoding/EncodingHelper.cs#L6105). It will absolutely fail when we enable 4:2:2 and 4:4:4. This is what we are going to do in JF 10.10/11, by adding a tool to detect the real hardware capabilities.

gnattu:
A format option is added for `scale_vt`. I limit the valid option to `nv12` and `p010` for now.

nyanmisaka:
> A format option is added for `scale_vt`. I limit the valid option to `nv12` and `p010` for now.

https://github.com/FFmpeg/FFmpeg/blob/64634e809f2e7b6c1a1ea3f6952a17c5915c3f22/libavcodec/videotoolbox.c#L1139-L1181

Can I assume that VtTransfer and VtDecoder share all input and output formats? If so I guess they could be added as well.

gnattu:
> Can I assume that VtTransfer and VtDecoder share all input and output formats? If so I guess they could be added as well.

The `VTPixelTransferSession` may not support all conversions to/from the formats supported by VideoToolbox, and NV12 and P010 are confirmed to work as the "safest" options. The encoder would also produce problematic outputs when a supported but not ideal input is given, just like the BGRA case we were seeing. If we want to add pixel formats that we are not sure will work, we can simply skip such checks altogether and let VideoToolbox error out if necessary. However, this approach cannot cover the case where the encoder produces wrong outputs.

nyanmisaka:
> > Can I assume that VtTransfer and VtDecoder share all input and output formats? If so I guess they could be added as well.
> 
> The `VTPixelTransferSession` may not support all conversions to/from the formats supported by VideoToolbox, and NV12 and P010 are confirmed to work as the "safest" options. The encoder would also produce problematic outputs when a supported but not ideal input is given, just like the BGRA case we were seeing. If we want to add pixel formats that we are not sure will work, we can simply skip such checks altogether and let VideoToolbox error out if necessary. However, this approach cannot cover the case where the encoder produces wrong outputs.

Is it possible to check `nv12` and `p010le` only when the user sets a different output format? (making it slightly relaxed just like upstream) If it's just scaling and no format conversion involved, I *think* it should be safe. (To make it not a breaking change)

gnattu:
> Is it possible to check nv12 and p010le only when the user sets a different output format? 

I think it already behaves like this?

https://github.com/gnattu/jellyfin-ffmpeg/blob/ac2e8120c170b1668470d348e7a842049adea315/debian/patches/0061-backport-scale-vt.patch#L308

```c
out_format = (s->format == AV_PIX_FMT_NONE) ? hw_frame_ctx_in->sw_format : s->format;
```

When the user sets no format, it would be `AV_PIX_FMT_NONE`, and in this case, it will just use whatever format is passed in.

nyanmisaka:
```
    out_format = (s->format == AV_PIX_FMT_NONE) ? hw_frame_ctx_in->sw_format : s->format;
    if (!format_is_supported(out_format)) {
        av_log(s, AV_LOG_ERROR, "Unsupported output format: %s\n",
               av_get_pix_fmt_name(out_format));
        return AVERROR(ENOSYS);
    }
```

//hw_frame_ctx_in->sw_format == p210le
//format_is_supported(hw_frame_ctx_in->sw_format) == 0

Currently, even if the user does not set the format option, an error will be returned when the input of `scale_vt` is non-nv12/p010le. There is no such restriction upstream.

gnattu:
Format check relaxed.

nyanmisaka (approved):

PR #353, opened 2024-03-09
Update CI to work with new repository layout

joshuaboniface:
**Changes**
1. Use new path layout.
2. Add major version calculation for use in path.
3. Remove sha256sums.
4. Drop non-LTS Ubuntus from builds (matches main server builds)

**Issues**
N/A

### DIFF
--- a/.github/workflows/_meta.yaml
+++ b/.github/workflows/_meta.yaml
@@ -104,7 +104,7 @@ jobs:
           key: ${{ secrets.REPO_KEY }}
           source: artifact/*
           strip_components: 1
-          target: /srv/incoming/ffmpeg/${{ steps.set_version.outputs.no-v }}/${{ inputs.distro }}/${{ matrix.arch }}/
+          target: /srv/incoming/ffmpeg/${{ steps.set_version.outputs.no-v }}/${{ inputs.distro }}/${{ matrix.arch }}
 
       - name: Move incoming release into repository
         uses: appleboy/ssh-action@8f949198563a347a01c65ffc60399aef2b59d4ab # v1.0.1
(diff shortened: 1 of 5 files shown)


nyanmisaka:
There are two more leftovers:
https://github.com/jellyfin/jellyfin-ffmpeg/blob/a1e482c4174d7d99d3769028c92b70bb47d780c4/.github/workflows/publish.yaml#L80-L81

nyanmisaka:
The [jellyfin-5.1](https://github.com/jellyfin/jellyfin-ffmpeg/tree/jellyfin-5.1) branch may also need to do the same thing, and we will need to maintain it for a while for backward compatibility reasons.

Shadowghost (approved):
Should work as intended.

joshuaboniface:
Done and done.

I see the MacOS build is failing, but I'm not sure if that's expected or not. Otherwise, hopefully all good!

nyanmisaka (approved):

nyanmisaka:
@joshuaboniface FYI we will defer adding the noble(ubu2404) build target because upstream Debian is transitioning to glib2.0, which results in some dev libs being unavailable in the non-proposed repo. Hopefully they will be available around the time ubu2404 is released.

- https://packages.qa.debian.org/g/glib2.0.html
- https://release.debian.org/transitions/html/auto-glib2.0.html
- https://packages.ubuntu.com/source/noble/harfbuzz  (gir1.2-*-2.0-dev not available)

joshuaboniface:
Sounds good @nyanmisaka yea I avoided adding it yet for the server CI too because the .NET repo isn't available for it yet :D

PR #355, opened 2024-03-09
Include dsc and tar archive in output

joshuaboniface:
Allows these to be uploaded to the repository server as well and be fetched from the repo.


joshuaboniface:
Closing as this turns out to be quite impossible without a major refactoring of how our ffmpeg is built.

PR #363, opened 2024-03-20
Fix secrets to work how it did before

joshuaboniface:
**Changes**
This failed, and I don't understand why it didn't work, but fall back.

**Issues**
N/A


gnattu (approved):
yolo

Shadowghost:
This failed because those vars originate from the input of the template workflow, they'd need to be changed in the workflow calling the template

PR #369, opened 2024-03-23
avfilter: add vf_tonemap_videotoolbox

### LINKED ISSUE #357: Improve tone mapping performance on Apple Silicon
Currently, for the 10.9 release, we are using OpenCL as a fallback for Video Tone Mapping in addition to VideoToolbox's native mechanism to support Dolby Vision Profile 5 Videos. It works, but the performance is very underwhelming on Apple Silicon GPUs. A 24-core M1 Max GPU can barely tone map 4K30 DoVi P5 videos.

One problem is the use of hardware surface. The current hardware mapping route between OpenCL and VideoToolbox is very slow. The `hwmap` and reverse `hwmap` chain significantly slows down the operation on 4K videos. By disabling the hardware surface and using `hwupload` and `hwdownload`, we could boost performance by almost 80% on a 4K video (52fps vs. 29fps).

The other problem stems from the GPU architecture itself and Apple's deprecation of OpenCL. The Apple GPU was designed to have slower communication between SIMD groups in a thread group but faster communication within a single SIMD. This tradeoff improves power efficiency and requires modifying algorithms to utilize SIMD-scoped operations. In OpenCL, Apple does not expose such operations to the shading language because they officially deprecated OpenCL, and this restriction fundamentally makes some OpenCL code bases slower than if written in Metal, including our OpenCL tone mapping kernels. It also makes matrix multiplication drop to 1/3 of maximum performance, becoming almost no faster than the CPU's AMX. By using libplacebo over the low-overhead MoltenVK, we can get another 80% performance boost (93 vs. 52) by using `hwupload`/`hwdownload`. The hwmap route is not implemented yet, but such performance improvement is already significant because this means with a powerful enough GPU, libplacebo can outperform VideoToolbox's native tone-mapping at the cost of higher power consumption.

To make this happen, we need to figure out a way to actually ship these binaries. ffmpeg loads Vulkan libs by `dlopen`, which means we need to ship `libMoltenVK.dylib` or let the user install it. Our portable builds will still be "portable" because they would run without MoltenVK, but they just cannot use libplacebo, which is in a similar situation as a Linux machine with a not correctly configured GPU driver. The configuration would be much simpler than configuring a Linux GPU, though. Just placing the `libMoltenVK.dylib` into the same folder as `ffmpeg` would be enough.

The next thing would be trying to implement a hwmap route to see if it improves performance further, but at this point, I have no confidence in `videotoolbox_vld` hardware surface. It is already broken in all our current use cases, and I don't expect it would work properly with Vulkan either, but I will try it.

I plan to migrate to libplacebo in the 10.10 release, and maybe implement some workaround to the current OpenCL filter on Apple silicon, like disabling hardware surface and/or force a quality tradeoff.

gnattu:
This filter does HDR(HDR10/HLG/Dolby Vision) to SDR conversion with tone-mapping using Metal.

For example:

```
ffmpeg \
  -init_hw_device videotoolbox \
  -i INPUT \
  -vf hwupload,tonemap_videotoolbox=t=bt2020:tonemap=linear:format=p010 \
  -c:v h264_videotoolbox \
  -b:v 2000k \
  -c:a copy \
  -y OUTPUT
```

**Changes**

**Issues**

Fixes #357

### DIFF
--- a/debian/patches/0069-add-tonemap-videotoolbox.patch
+++ b/debian/patches/0069-add-tonemap-videotoolbox.patch
@@ -12,7 +12,7 @@ diff --git a/configure b/configure
  tonemap_filter_deps="const_nan"
  tonemap_vaapi_filter_deps="vaapi VAProcFilterParameterBufferHDRToneMapping"
  tonemap_opencl_filter_deps="opencl const_nan"
-+tonemap_videotoolbox_filter_deps="metal corevideo videotoolbox"
++tonemap_videotoolbox_filter_deps="metal corevideo videotoolbox const_nan"
  transpose_opencl_filter_deps="opencl"
  transpose_vaapi_filter_deps="vaapi VAProcPipelineCaps_rotation_flags"
  transpose_vulkan_filter_deps="vulkan spirv_compiler"
@@ -1523,7 +1523,7 @@ new file mode 100644
 +    out_frames_ctx->width = outlink->w;
 +    out_frames_ctx->height = outlink->h;
 +
-+    ret = ff_filter_init_hw_frames(avctx, outlink, 10);
++    ret = ff_filter_init_hw_frames(avctx, outlink, 1);
 +    if (ret < 0)
 +        return ret;
 +
--- a/debian/patches/0069-add-tonemap-videotoolbox.patch
+++ b/debian/patches/0069-add-tonemap-videotoolbox.patch
@@ -1,0 +1,1867 @@
+Subject: [PATCH] avfilter: add vf_tonemap_videotoolbox
+---
+Index: configure
+IDEA additional info:
+Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
+<+>UTF-8
+===================================================================
+diff --git a/configure b/configure
+--- a/configure	(revision 2e3649a10ff797e360eb9fd5d389dfa6611dea59)
++++ b/configure	(revision bcb63879b775fef8085a5311f045dc1cc4960a98)
+@@ -3748,6 +3748,7 @@
+ tonemap_filter_deps="const_nan"
+ tonemap_vaapi_filter_deps="vaapi VAProcFilterParameterBufferHDRToneMapping"
+ tonemap_opencl_filter_deps="opencl const_nan"
++tonemap_videotoolbox_filter_deps="metal corevideo videotoolbox"
+ transpose_opencl_filter_deps="opencl"
+ transpose_vaapi_filter_deps="vaapi VAProcPipelineCaps_rotation_flags"
+ transpose_vulkan_filter_deps="vulkan spirv_compiler"
+Index: libavfilter/Makefile
+IDEA additional info:
+Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
+<+>UTF-8
+===================================================================
+diff --git a/libavfilter/Makefile b/libavfilter/Makefile
+--- a/libavfilter/Makefile	(revision 2e3649a10ff797e360eb9fd5d389dfa6611dea59)
++++ b/libavfilter/Makefile	(revision bcb63879b775fef8085a5311f045dc1cc4960a98)
+@@ -510,6 +510,9 @@
+ OBJS-$(CONFIG_TONEMAP_CUDA_FILTER)           += vf_tonemap_cuda.o cuda/tonemap.ptx.o \
+                                                 cuda/host_util.o
+ OBJS-$(CONFIG_TONEMAP_VAAPI_FILTER)          += vf_tonemap_vaapi.o vaapi_vpp.o
++OBJS-$(CONFIG_TONEMAP_VIDEOTOOLBOX_FILTER)   += vf_tonemap_videotoolbox.o \
++                                                metal/vf_tonemap_videotoolbox.metallib.o \
++                                                metal/utils.o
+ OBJS-$(CONFIG_TPAD_FILTER)                   += vf_tpad.o
+ OBJS-$(CONFIG_TRANSPOSE_FILTER)              += vf_transpose.o
+ OBJS-$(CONFIG_TRANSPOSE_CUDA_FILTER)         += vf_transpose_cuda.o vf_transpose_cuda.ptx.o \
+                                                 cuda/load_helper.o
+Index: libavfilter/allfilters.c
+IDEA additional info:
+Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
+<+>UTF-8
+===================================================================
+diff --git a/libavfilter/allfilters.c b/libavfilter/allfilters.c
+--- a/libavfilter/allfilters.c	(revision 2e3649a10ff797e360eb9fd5d389dfa6611dea59)
++++ b/libavfilter/allfilters.c	(revision bcb63879b775fef8085a5311f045dc1cc4960a98)
+@@ -480,6 +480,7 @@
+ extern const AVFilter ff_vf_tonemap_cuda;
+ extern const AVFilter ff_vf_tonemap_opencl;
+ extern const AVFilter ff_vf_tonemap_vaapi;
++extern const AVFilter ff_vf_tonemap_videotoolbox;
+ extern const AVFilter ff_vf_tpad;
+ extern const AVFilter ff_vf_transpose;
+ extern const AVFilter ff_vf_transpose_cuda;
+Index: libavfilter/metal/vf_tonemap_videotoolbox.metal
+IDEA additional info:
+Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
+<+>UTF-8
+===================================================================
+diff --git a/libavfilter/metal/vf_tonemap_videotoolbox.metal b/libavfilter/metal/vf_tonemap_videotoolbox.metal
+new file mode 100644
+--- /dev/null	(revision bcb63879b775fef8085a5311f045dc1cc4960a98)
++++ b/libavfilter/metal/vf_tonemap_videotoolbox.metal	(revision bcb63879b775fef8085a5311f045dc1cc4960a98)
+@@ -0,0 +1,688 @@
++/*
++ * Copyright (c) 2024 Gnattu OC <gnattuoc@me.com>
++ *
++ * This file is part of FFmpeg.
++ *
++ * FFmpeg is free software; you can redistribute it and/or
++ * modify it under the terms of the GNU Lesser General Public
++ * License as published by the Free Software Foundation; either
++ * version 2.1 of the License, or (at your option) any later version.
++ *
++ * FFmpeg is distributed in the hope that it will be useful,
++ * but WITHOUT ANY WARRANTY; without even the implied warranty of
++ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
++ * Lesser General Public License for more details.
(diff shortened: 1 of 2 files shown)


gnattu:
A lot of the context used in this file is missing in the upstream. Is it possible to have this upstreamed?

nyanmisaka:
> A lot of the context used in this file is missing in the upstream. Is it possible to have this upstreamed?

The existence of vf_libplacebo makes it more difficult to upstream new GPGPU-based tonemap filters. They may require your filter to behave like libplacebo, even though ours is a stripped-down version. In addition, new DoVi-related structures and funcs need to be discussed and defined in colorspace.h. Finally, there are very few developers in upstream who can review Metal/Obj-C code, and such huge filters may be ignored for months or even forever. . .

debian/patches/0069-add-tonemap-videotoolbox.patch
```diff
@@ -0,0 +1,1867 @@
...
+<+>UTF-8
+===================================================================
+diff --git a/configure b/configure
+--- a/configure	(revision 2e3649a10ff797e360eb9fd5d389dfa6611dea59)
++++ b/configure	(revision bcb63879b775fef8085a5311f045dc1cc4960a98)
+@@ -3748,6 +3748,7 @@
+ tonemap_filter_deps="const_nan"
+ tonemap_vaapi_filter_deps="vaapi VAProcFilterParameterBufferHDRToneMapping"
+ tonemap_opencl_filter_deps="opencl const_nan"
++tonemap_videotoolbox_filter_deps="metal corevideo videotoolbox"
```
nyanmisaka:
```diff
+        case TONEMAP_GAMMA:
+            if (isnan(ctx->param))
+                ctx->param = 1.8f;
```

`isnan()` depends on `const_nan`.

debian/patches/0069-add-tonemap-videotoolbox.patch
```diff
@@ -0,0 +1,1867 @@
...
++#define pivots_sz pivots_cnt*sizeof(float)
++#define coeffs_sz coeffs_cnt*sizeof(float)
++#define mmr_sz mmr_cnt*sizeof(float)
++
++extern char ff_vf_tonemap_videotoolbox_metallib_data[];
++extern unsigned int ff_vf_tonemap_videotoolbox_metallib_len;
++
++static const enum AVPixelFormat supported_formats[] = {
++    AV_PIX_FMT_NV12,
++    AV_PIX_FMT_P010,
```
nyanmisaka, line 812: Missing format that matches `in_planes > 2` and `out_planes > 2`.
Is this intended? Or they are just not tested?

debian/patches/0069-add-tonemap-videotoolbox.patch
```diff
@@ -0,0 +1,1867 @@
...
++        av_log(ctx, AV_LOG_ERROR, "Unsupported input format: %s\n",
++               av_get_pix_fmt_name(in_format));
++        return AVERROR(ENOSYS);
++    }
++    if (!format_is_supported(out_format)) {
++        av_log(ctx, AV_LOG_ERROR, "Unsupported output format: %s\n",
++               av_get_pix_fmt_name(out_format));
++        return AVERROR(ENOSYS);
++    }
++    if (in_desc->comp[0].depth != 10 && in_desc->comp[0].depth != 16) {
```
nyanmisaka, line 1503: P016?

debian/patches/0069-add-tonemap-videotoolbox.patch
```diff
@@ -0,0 +1,1867 @@
...
++    outlink->hw_frames_ctx = av_hwframe_ctx_alloc(in_frames_ctx->device_ref);
++    outlink->w = inlink->w;
++    outlink->h = inlink->h;
++    out_frames_ctx = (AVHWFramesContext *)outlink->hw_frames_ctx->data;
++    out_frames_ctx->format = AV_PIX_FMT_VIDEOTOOLBOX;
++    out_frames_ctx->sw_format = out_format;
++    out_frames_ctx->width = outlink->w;
++    out_frames_ctx->height = outlink->h;
++
++    ret = ff_filter_init_hw_frames(avctx, outlink, 10);
```
nyanmisaka: 1 should be enough.

debian/patches/0069-add-tonemap-videotoolbox.patch
gnattu: These formats have not been tested, so I conservatively removed them. The code path for non-biplanar formats is kept just in case they can be used.

debian/patches/0069-add-tonemap-videotoolbox.patch
gnattu: As mentioned above, this is a "just-in-case" option. I don't have any video with this format to test, so I disabled it in the `supported_formats` section, but the code path is kept.

nyanmisaka:
Here are [some clips](https://drive.google.com/drive/folders/127DXU2hvmzXgbzspmhMDUoQXRnGi3yqU?usp=sharing) I collected for testing DoVi/HDR. You can play it to see if the color is normal and whether it will cause ffmpeg to fail.

gnattu:
> Here are [some clips](https://drive.google.com/drive/folders/127DXU2hvmzXgbzspmhMDUoQXRnGi3yqU?usp=sharing) I collected for testing DoVi/HDR. You can play it to see if the color is normal and whether it will cause ffmpeg to fail.

These are all P010 videos, and I've already tested a lot of them. I'm leaving non P010 and NV12 formats off until we have samples that confirmed to work.

nyanmisaka:
> > Here are [some clips](https://drive.google.com/drive/folders/127DXU2hvmzXgbzspmhMDUoQXRnGi3yqU?usp=sharing) I collected for testing DoVi/HDR. You can play it to see if the color is normal and whether it will cause ffmpeg to fail.
> 
> These are all P010 videos, and I've already tested a lot of them. I'm leaving non P010 and NV12 formats off until we have samples that confirmed to work.

What I mean is that it contains clips that need to reset the filter. As for non-(4:2:0 10-bit) DoVi, I think it won't happen any time soon.

gnattu:
> What I mean is that it contains clips that need to reset the filter. 

I tested all of them and every sample looks good.

> As for non-(4:2:0 10-bit) DoVi, I think it won't happen any time soon.

Then keep p010 and nv12 as only supported formats for now.

One interesting thing is, during my testing I found an ffmpeg issue on `bufsize` and `maxrate` option, where setting them will cause performance regression and even encoder hangs when the video bitrate is high enough. I'm going to push a server side change as well.

nyanmisaka:
> One interesting thing is, during my testing I found an ffmpeg issue on `bufsize` and `maxrate` option, where setting them will cause performance regression and even encoder hangs when the video bitrate is high enough. I'm going to push a server side change as well.

These are generic lavc options. Encoders should adapt to these options and throw an error if the user settings exceed the hardware specs, or automatically roll back to the default values.

gnattu:
It's not about "exceeding hardware specs"; it's a supported value, but it just works buggy. The encoder would succeed, albeit with a much slower speed (up to a 50% regression), and in some circumstances, the ffmpeg process just hangs and doesn't proceed. This happens randomly. Removing the `maxrate` and `bufsize` , set only `-b:v`(`-b:v` and `maxrate` are the same value in this case, and `bufsize=maxrate *2` like what we are doing in jellyfin) and letting VideoToolbox handle it automatically mitigates this issue.

gnattu:
```
2024-03-23T15:51:40.8291924Z + git clone -b stripped4 --depth=1 https://gitlab.freedesktop.org/wtaymans/fdk-aac-stripped.git
2024-03-23T15:51:40.8305011Z Cloning into 'fdk-aac-stripped'...
2024-03-23T15:54:34.6733077Z error: RPC failed; HTTP 502 curl 22 The requested URL returned error: 502
```

CI failure due to remote issue

nyanmisaka:
> It's not about "exceeding hardware specs"; it's a supported value, but it just works buggy. The encoder would succeed, albeit with a much slower speed (up to a 50% regression), and in some circumstances, the ffmpeg process just hangs and doesn't proceed. This happens randomly. Removing the `maxrate` and `bufsize` , set only `-b:v`(`-b:v` and `maxrate` are the same value in this case, and `bufsize=maxrate *2` like what we are doing in jellyfin) and letting VideoToolbox handle it automatically mitigates this issue.

Ideally this undefined behavior should not occur. Either work or fail. If removing them does not cause rate control regression, then there is no problem.
```
// not supported
bufsize(avctx->rc_buffer_size)

// https://github.com/FFmpeg/FFmpeg/blob/bfbf0f4e82eff7fc6f15205b71e97322e7681dcb/libavcodec/videotoolboxenc.c#L1273-L1318
maxrate(avctx->rc_max_rate)
```

gnattu:
>  If removing them does not cause rate control regression, then there is no problem.

VT obeys the `b:v` option quite well and would not cause the bitrate to be much higher than that, so the `maxrate` can be safely removed.

My suggested changes:

https://github.com/jellyfin/jellyfin/pull/11198/commits/2933ec2eb7e409b295e67e0e4333b161f9a00726

debian/patches/0069-add-tonemap-videotoolbox.patch
```diff
@@ -0,0 +1,1867 @@
...
++
++void map_four_pixels_rgb(thread float4 *r4, thread float4 *g4, thread float4 *b4, float peak) {
++#define MAP_FOUR_PIXELS(sig, peak, target_peak) \
++{ \
++    sig.x = tonemap(sig.x, peak, target_peak); \
++    sig.y = tonemap(sig.y, peak, target_peak); \
++    sig.z = tonemap(sig.z, peak, target_peak); \
++    sig.w = tonemap(sig.w, peak, target_peak); \
++}
++    if (is_tone_mode_rgb) {
```
nyanmisaka, line 461: Will this runtime if/else check cause branch divergence on Metal?
I know Cuda doesn't like this very much, so I have to create kernels for every path.

gnattu: This isn't a runtime if-else check after the Metal magic. The `is_tone_mode_rgb` is referred to as *function constants* and is a value to be set at runtime but stays constant during each function calls. When we are creating the Metal function with function constants, the Metal backend compiler will compile the specialized function at runtime, eliminating all constant branches with `if(false)`. When this is uploaded to the GPU, only the branch evaluated to true will remain.
https://github.com/gnattu/jellyfin-ffmpeg/blob/9b2975da5170275d5715c4be7e3367432a4a1729/debian/patches/0069-add-tonemap-videotoolbox.patch#L1433

nyanmisaka: Then it seems that Cuda doesn't do link-time optimization (JIT LTO). cuLink (constants ptx + tonemap ptx), because ptx compilation occurs when ffmpeg is compiled.

So, when I use `__constant__ cond` for if(cond)...else in Cuda, there is still a performance penalty compared to no branching.

gnattu: On Apple Silicon Macs the performance is currently limited by the hardware decoder/encoders or whatever throttling Apple is putting on it. Even removing the tone-map filter altogether the fps does not increase, so our kernel is doing its best now.

nyanmisaka (approved):
yolo

PR #381, opened 2024-04-21
fix: remove fuzz

gnattu:
**Changes**

The mjpeg patch now contains fuzz after merging, remove it.

**Issues**


nyanmisaka (approved):
https://github.com/jellyfin/jellyfin-ffmpeg/blob/3c9bf864d38fa52f82a6523b9dfe45e933df23e6/debian/changelog#L1-L4

Add this to changelog?

Shadowghost (approved):

PR #388, opened 2024-05-07
lavc/videotoolbox: prepare for trickplay

gnattu:
**Changes**

- Enable keyframe only decoding for videotoolbox for faster trickplay image generation 
- Add low priority decoding support for videotoolbox to be used with trickplay
- Correctly map mjpeg_videotoolbox's allow_sw and format options

**Issues**


Shadowghost (approved):

debian/patches/0074-vt-enable-keyframe-only-decoding.patch
```diff
@@ -0,0 +1,19 @@
...
+--- a/libavcodec/videotoolbox.c
++++ b/libavcodec/videotoolbox.c
+@@ -983,6 +983,14 @@ static int videotoolbox_start(AVCodecContext *avctx)
+             av_log(avctx, AV_LOG_ERROR, "VideoToolbox reported invalid data.\n");
+             return AVERROR_INVALIDDATA;
+         case 0:
++            if (avctx->skip_frame >= AVDISCARD_NONKEY) {
++                status = VTSessionSetProperty(videotoolbox->session,
++                                              kVTDecompressionPropertyKey_OnlyTheseFrames,
++                                              kVTDecompressionProperty_OnlyTheseFrames_KeyFrames);
```
nyanmisaka: What would happen without this videotoolbox setting?

gnattu: It would still work, this is added to make sure the decoder works as intended.

nyanmisaka (approved):

debian/patches/0074-vt-low-priority-keyframe-decoding.patch
```diff
@@ -0,0 +1,72 @@
...
++                                              kVTDecompressionProperty_OnlyTheseFrames_KeyFrames);
++                if (status) {
++                    av_log(avctx, AV_LOG_WARNING, "kVTDecompressionProperty_OnlyTheseFrames_KeyFrames is not supported on this device. Ignoring.\n");
++                }
++            }
++            if (avctx->hwaccel_flags & AV_HWACCEL_FLAG_LOW_PRIORITY) {
++                status = VTSessionSetProperty(videotoolbox->session,
++                                              kVTDecompressionPropertyKey_RealTime,
++                                              kCFBooleanFalse);
++                av_log(avctx, AV_LOG_INFO, "Decoder running at lower priority.\n");
```
nyanmisaka: Will enabling it hard limit the decoding speed of this session? Or is it dynamically adjusted based on foreground tasks?

gnattu: No, from my testing it does not set hard limit. The image generation speed does not change if the decoder is not busy. It only place this session at a lower priority so that it will be throttled first when the decoder is overloaded.

nyanmisaka (approved):

nyanmisaka:
Do you want to let this in the next tag, or wait for 10.10?

gnattu:
I pushed with rebasing on the latest branch. This can be released with the next tag, which will also help to reduce merge conflicts.

nyanmisaka (approved):

PR #397, opened 2024-06-02
Add detection of DTS:X and DTS:X IMAX

Bond-009:
Backport of https://git.ffmpeg.org/gitweb/ffmpeg.git/commitdiff/98a4699216fbc3a4e382ba6a10026eab39b5214e

### DIFF
--- a/debian/patches/0077-add-detection-of-dtsx.patch
+++ b/debian/patches/0077-add-detection-of-dtsx.patch
@@ -1,0 +1,128 @@
+Index: jellyfin-ffmpeg/libavcodec/avcodec.h
+===================================================================
+--- jellyfin-ffmpeg.orig/libavcodec/avcodec.h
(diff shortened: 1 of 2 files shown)


nyanmisaka (approved):
Are any subsequent changes required on the server side?

Shadowghost (approved):

Bond-009:
Server changes are already done

https://github.com/jellyfin/jellyfin/pull/11772

gnattu:
So this is for 10.10 right?

Shadowghost:
Yes, it was merged to master after we cut 10.9.0

mohd-akram:
> So this is for 10.10 right?

IMO this and #399 should be backported to 10.9 due to jellyfin/jellyfin#9996.

Also using it in jellyfin/jellyfin-androidtv#2893.

PR #404, opened 2024-06-22
[Do not merge] test dovi tonemapx

gnattu:
**Changes**

**Issues**

### DIFF
--- a/debian/patches/0081-dovi-tonemapx.patch
+++ b/debian/patches/0081-dovi-tonemapx.patch
@@ -1,0 +1,446 @@
+Index: FFmpeg/libavfilter/vf_tonemapx.c
+===================================================================
+--- FFmpeg.orig/libavfilter/vf_tonemapx.c
++++ FFmpeg/libavfilter/vf_tonemapx.c
+@@ -89,6 +89,7 @@ typedef struct TonemapxContext {
+     float *tonemap_lut;
+     uint16_t *delin_lut;
+     int in_yuv_off, out_yuv_off;
++    struct DoviMetadata *dovi;
+ 
+     DECLARE_ALIGNED(16, int16_t, yuv2rgb_coeffs)[3][3][8];
+     DECLARE_ALIGNED(16, int16_t, rgb2yuv_coeffs)[3][3][8];
+@@ -113,6 +114,13 @@ static const enum AVPixelFormat out_pix_
+     AV_PIX_FMT_P016,
+ };
+ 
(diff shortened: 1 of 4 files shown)


debian/patches/0081-dovi-tonemapx.patch
```diff
@@ -0,0 +1,444 @@
...
++            ycc2rgb(c1, yuv1[0], yuv1[1], yuv1[2], s->dovi->nonlinear, s->dovi->nonlinear_offset);
++            ycc2rgb(c2, yuv2[0], yuv2[1], yuv2[2], s->dovi->nonlinear, s->dovi->nonlinear_offset);
++            ycc2rgb(c3, yuv3[0], yuv3[1], yuv3[2], s->dovi->nonlinear, s->dovi->nonlinear_offset);
++            ycc2rgb(c4, yuv4[0], yuv4[1], yuv4[2], s->dovi->nonlinear, s->dovi->nonlinear_offset);
++            lms2rgb(c1, c1[0], c1[1], c1[2], s->dovi->linear);
++            lms2rgb(c2, c2[0], c2[1], c2[2], s->dovi->linear);
++            lms2rgb(c3, c3[0], c3[1], c3[2], s->dovi->linear);
++            lms2rgb(c4, c4[0], c4[1], c4[2], s->dovi->linear);
++
++            r[0] = av_clip_int16(c1[0] * 28672);
```
nyanmisaka, line 282: This will make the picture look slightly correct, but it won't solve the problem.
`r[0] = av_clip_int16((int)(c1[0] * 28672) >> in_sh);`

In theory, this will give you the normalized pixel value for p010 input.
`pix_val / ((1<<10)-1) => 0.0f~1.0f`

But anyway the result of reshaped rgb output is wrong (I commented out tonemap_int16() to check)

This is really weird. I'll check it out again when I have some free time.

gnattu: I have investigated this a little bit more and found that:

For yuv420p10 inputs, we need to divide by 1023, to get the normalization, but then we convert it back we need to multiple by 65535 to reinterpret it as a full 16bit color. the yuv420p10 inputs does not need to be shifted to low 10bits.

Current reshaping implementation seems to only reshape the intensity correctly and the chromas are not correctly reshaped. Doing `CLAMP(s, comp->pivots[0], comp->pivots[comp->num_pivots-1]);` on chromas will make all chromas to be a very small number (and mostly zero), the resulting video is mostly black-white. Not doing this the output video will behave like having a blue mask on it.

PR #405, opened 2024-06-24
avfilter/colorspace: fix dovi peak calculation

gnattu:
The max PQ to max mastering display luminance calculation is wrong and not within the scale of ffmpeg. This will overflow the pixels during tone mapping unless the user supplies a peak value to overwrite this.

- No need to multiply by reference white for PQ to luminance
- The ffmpeg brightness scale for SMPTE ST.2084 is 100 not 10000

Before: 811943.437500
After: 399.9721366995 (4000 nit master display)

**Changes**

**Issues**


Shadowghost (approved):

gnattu:
ubuntu noble fails again, not related to this particular PR

Bond-009 (approved):

nyanmisaka (approved):

PR #407, opened 2024-06-25
avfilter/tonemapx: add simd optimized tonemapx

gnattu:
This includes NEON for ARMv8, SSE for x86-64-v2 and AVX+FMA for x86-64-v3

Test result with 4K HEVC 10bit HLG input, encoding with libx264 veryfast using bt2390:

Intel Core i9-12900:

tonemapx.c: 57fps
tonemapx.sse: 74fps
tonemapx.avx: 77fps

Apple M1 Max:

tonemapx.c:43fps
tonemapx.neon: 57fps

For comparison, original zscale+tonemap simd results:

Intel Core i9-12900:

tonemap.avx: 40fps
tonemap.sse: 40fps
tonemap.c: 32fps

Apple M1 Max:

tonemap.neon: 44fps
tonemap.c: 35fps

The original implementation is too memory heavy that dual-channel desktop CPUs are easily memory bounded due to the intermediate RGBF32 framebuffer sharing with zscale. Tonemapx lowered the the bandwidth requirement which brings significant performance gain to bandwidth limited platforms. Even for bandwidth-rich M1 Max it still provides significant performance boost due to better cache hitrate.

**Changes**

- Fixed an issue that the input parameter is not really initialized
- Added neon, avx and sse code path
- Use function pointer instead of branching in each thread for tonemap function selection

**Issues**

Replaces #401

### DIFF
--- a/builder/variants/defaults-gpl.sh
+++ b/builder/variants/defaults-gpl.sh
@@ -6,5 +6,3 @@ GIT_BRANCH="jellyfin"
 LICENSE_FILE="COPYING.GPLv3"
 
 [[ $TARGET == linux* ]] && FF_CONFIGURE+=" --disable-libxcb --disable-xlib" || true
-[[ $TARGET == win* ]] && FF_CFLAGS+=" -Wa,-muse-unaligned-vector-move" || true
-[[ $TARGET == win* ]] && FF_CXXFLAGS+=" -Wa,-muse-unaligned-vector-move" || true
--- a/builder/variants/defaults-gpl.sh
+++ b/builder/variants/defaults-gpl.sh
@@ -6,3 +6,5 @@ GIT_BRANCH="jellyfin"
 LICENSE_FILE="COPYING.GPLv3"
 
 [[ $TARGET == linux* ]] && FF_CONFIGURE+=" --disable-libxcb --disable-xlib" || true
+[[ $TARGET == win* ]] && FF_CFLAGS+=" -Wa,-muse-unaligned-vector-move" || true
+[[ $TARGET == win* ]] && FF_CXXFLAGS+=" -Wa,-muse-unaligned-vector-move" || true
--- a/debian/patches/0080-add-tonemapx-filter.patch
+++ b/debian/patches/0080-add-tonemapx-filter.patch
@@ -115,7 +115,7 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/vf_tonemapx.c
-@@ -0,0 +1,3153 @@
+@@ -0,0 +1,3151 @@
 +/*
 + * This file is part of FFmpeg.
 + *
@@ -154,10 +154,8 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 +#if  ARCH_AARCH64
 +#   include <arm_neon.h>
 +#   include "libavutil/aarch64/cpu.h"
-+#define __builtin_ia32_extract128i256(a, b) _mm_setzero_si128()
-+#define __builtin_ia32_permdi256(a, b) _mm256_setzero_si256()
 +#endif
-+#if ARCH_X86 || ARCH_AARCH64
++#if ARCH_X86
 +#   include <immintrin.h>
 +#   include <emmintrin.h>
 +#   include <smmintrin.h>
@@ -586,7 +584,7 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 +    *b_out = delin_lut[av_clip_uintp2(b_lin * 32767 + 0.5, 15)];
 +}
 +
-+#if ARCH_X86 || ARCH_AARCH64
++#if ARCH_X86
 +X86_64_V2 static inline __m128i av_clip_uint16_sse(__m128i a)
 +{
 +    __m128i mask = _mm_set1_epi32(0x7FFF);
@@ -1122,7 +1120,7 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 +    }
 +}
 +
-+#if ARCH_X86 || ARCH_AARCH64
++#if ARCH_X86
 +X86_64_V2 static inline __m128i av_clip_int16_sse(__m128i a)
 +{
 +    __m128i add_result = _mm_add_epi32(a, _mm_set1_epi32(0x8000U));
@@ -2103,7 +2101,7 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 +    }
 +}
 +
-+#if ARCH_X86 || ARCH_AARCH64
++#if ARCH_X86
 +X86_64_V2 static void tonemap_frame_p016_p010_2_p016_p010_sse(uint16_t *dsty, uint16_t *dstuv,
 +                                                              const uint16_t *srcy, const uint16_t *srcuv,
 +                                                              const int *dstlinesize, const int *srclinesize,
--- a/debian/patches/0080-add-tonemapx-filter.patch
+++ b/debian/patches/0080-add-tonemapx-filter.patch
@@ -115,7 +115,7 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/vf_tonemapx.c
-@@ -0,0 +1,5147 @@
+@@ -0,0 +1,1203 @@
 +/*
 + * This file is part of FFmpeg.
 + *
@@ -140,46 +140,29 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 + */
 +
 +#include <float.h>
-+#include <stdio.h>
 +#include <string.h>
 +
 +#include "libavutil/avassert.h"
 +#include "libavutil/imgutils.h"
 +#include "libavutil/internal.h"
-+#include "libavutil/intreadwrite.h"
 +#include "libavutil/mem_internal.h"
 +#include "libavutil/opt.h"
-+#include "libavutil/pixdesc.h"
 +#include "libavutil/cpu.h"
 +#if  ARCH_AARCH64
-+#   include <arm_neon.h>
 +#   include "libavutil/aarch64/cpu.h"
++#   include "aarch64/vf_tonemapx_intrin_neon.h"
 +#endif
 +#if ARCH_X86
-+#   include <immintrin.h>
-+#   include <emmintrin.h>
-+#   include <smmintrin.h>
 +#   include "libavutil/x86/cpu.h"
++#   include "x86/vf_tonemapx_intrin_sse.h"
++#   include "x86/vf_tonemapx_intrin_avx.h"
 +#endif
 +
 +#include "avfilter.h"
-+#include "colorspace.h"
 +#include "formats.h"
 +#include "internal.h"
 +#include "video.h"
-+
-+#define REFERENCE_WHITE 203.0f
-+#define FLOAT_EPS 1.175494351e-38f
-+
-+#if defined(__GNUC__) || defined(__clang__)
-+#    if (__GNUC__ >= 11) || (__clang_major__ >= 12)
-+#        define X86_64_V2 __attribute__((target("arch=x86-64-v2")))
-+#        define X86_64_V3 __attribute__((target("arch=x86-64-v3")))
-+#    else
-+#        define X86_64_V2 __attribute__((target("sse4.2")))
-+#        define X86_64_V3 __attribute__((target("avx2,fma")))
-+#    endif // (__GNUC__ >= 11) || (__clang_major__ >= 12)
-+#endif // defined(__GNUC__) || defined(__clang__)
++#include "vf_tonemapx.h"
 +
 +enum TonemapAlgorithm {
 +    TONEMAP_NONE,
@@ -192,20 +175,6 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 +    TONEMAP_BT2390,
 +    TONEMAP_MAX,
 +};
-+
-+typedef struct TonemapIntParams {
-+    double lut_peak;
-+    float *lin_lut;
-+    float *tonemap_lut;
-+    uint16_t *delin_lut;
-+    int in_yuv_off, out_yuv_off;
-+    int16_t (*yuv2rgb_coeffs)[3][3][8];
-+    int16_t (*rgb2yuv_coeffs)[3][3][8];
-+    double  (*rgb2rgb_coeffs)[3][3];
-+    int rgb2rgb_passthrough;
-+    const AVLumaCoefficients *coeffs, *ocoeffs;
-+    double desat;
-+} TonemapIntParams;
 +
 +typedef struct TonemapxContext {
 +    const AVClass *class;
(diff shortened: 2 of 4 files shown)


nyanmisaka:
Got 74fps with the tonemapx.avx on 5950x on Windows. I'm quite pleased with the performance.

Further adding yuv420p10 input and yuv420p output support should squeeze out more FPS, because IIRC yuv420p1x to p01x conversion in FFmpeg is written in C, and other encoders such as libx265 and svt-av1 do not support nv12 input.

gnattu:
> Further adding yuv420p10 input and yuv420p output support should squeeze out more FPS, because IIRC yuv420p1x to p01x conversion in FFmpeg is written in C, and other encoders such as libx265 and svt-av1 do not support nv12 input.

Probably not that much for bandwidth rich systems as this conversion is mainly memory operation not compute operation, could be useful for bandwidth constrained platforms if we process yuv420p directly as the frame copy is further reduced and our avx implementation could see more improvements.

nyanmisaka:
> > Further adding yuv420p10 input and yuv420p output support should squeeze out more FPS, because IIRC yuv420p1x to p01x conversion in FFmpeg is written in C, and other encoders such as libx265 and svt-av1 do not support nv12 input.
> 
> Probably not that much for bandwidth rich systems as this conversion is mainly memory operation not compute operation, could be useful for bandwidth constrained platforms if we process yuv420p directly as the frame copy is further reduced and our avx implementation could see more improvements.

We really can't expect all our users to have i9/R9/Apple Silicon Max. FWIW in this simple test it already clearly hurts performance, although it wouldn't be so noticeable in a real world testing because this version is already much better than the zscale+tonemap.orig.

```
./ffmpeg -threads 16 -i 4K_HLG.MP4 -an -sn -vf format=yuv420p10 -vframes 1000 -f null -
```
`frame= 1000 fps=251 q=-0.0 Lsize=N/A time=00:00:16.71 bitrate=N/A speed=4.19x`

```
./ffmpeg -threads 16 -i 4K_HLG.MP4 -an -sn -vf format=yuv420p10,format=p010 -vframes 1000 -f null -
```
`frame= 1000 fps=168 q=-0.0 Lsize=N/A time=00:00:16.71 bitrate=N/A speed=2.81x`

gnattu:
> We really can't expect all our users to have i9/R9/Apple Silicon Max. FWIW in this simple test it already clearly hurts performance, although it wouldn't be so noticeable in a real world testing because this version is already much better than the zscale+tonemap.orig.
> 
> ```
> ./ffmpeg -threads 16 -i 4K_HLG.MP4 -an -sn -vf format=yuv420p10 -vframes 1000 -f null -
> ```
> 
> `frame= 1000 fps=251 q=-0.0 Lsize=N/A time=00:00:16.71 bitrate=N/A speed=4.19x`
> 
> ```
> ./ffmpeg -threads 16 -i 4K_HLG.MP4 -an -sn -vf format=yuv420p10,format=p010 -vframes 1000 -f null -
> ```
> 
> `frame= 1000 fps=168 q=-0.0 Lsize=N/A time=00:00:16.71 bitrate=N/A speed=2.81x`

Do you have time for a reference c implementation of yuv420p? I can port SIMD version to it.

nyanmisaka:
> > We really can't expect all our users to have i9/R9/Apple Silicon Max. FWIW in this simple test it already clearly hurts performance, although it wouldn't be so noticeable in a real world testing because this version is already much better than the zscale+tonemap.orig.
> > ```
> > ./ffmpeg -threads 16 -i 4K_HLG.MP4 -an -sn -vf format=yuv420p10 -vframes 1000 -f null -
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `frame= 1000 fps=251 q=-0.0 Lsize=N/A time=00:00:16.71 bitrate=N/A speed=4.19x`
> > ```
> > ./ffmpeg -threads 16 -i 4K_HLG.MP4 -an -sn -vf format=yuv420p10,format=p010 -vframes 1000 -f null -
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `frame= 1000 fps=168 q=-0.0 Lsize=N/A time=00:00:16.71 bitrate=N/A speed=2.81x`
> 
> Do you have time for a reference c implementation of yuv420p? I can port SIMD version to it.

Not yet. But here are some examples. Just convert it to use in a for loop. The difference between nv12/p01x and yuv420p/yuv420p1x is only whether U and V are interleaved.

https://github.com/jellyfin/jellyfin-ffmpeg/blob/44e1bf07446cb2a280ed55a08f303a1014ab5bfe/debian/patches/0005-add-cuda-tonemap-impl.patch#L745-L883

gnattu:
gnu.org is down for 9 hours now and failing the pipeline

nyanmisaka:
ffmpeg crashes after inserting a downscaling filter and running for a while.
`scale=s=1920x1080:flags=fast_bilinear,tonemapx=...`

For example, these resolutions.
1280x720
1920x1080
1920x1088
2560x1440

It can't be reproduced in tonemapx.c. I guess it has something to do with the edges of the image that cannot be accelerated by SIMD.

```
- <Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
- <System>
  <Provider Name="Application Error" /> 
  <EventID Qualifiers="0">1000</EventID> 
  <Version>0</Version> 
  <Level>2</Level> 
  <Task>100</Task> 
  <Opcode>0</Opcode> 
  <Keywords>0x80000000000000</Keywords> 
  <TimeCreated SystemTime="2024-06-26T02:48:40.9359314Z" /> 
  <EventRecordID>216441</EventRecordID> 
  <Correlation /> 
  <Execution ProcessID="0" ThreadID="0" /> 
  <Channel>Application</Channel> 
  <Computer>pc</Computer> 
  <Security /> 
  </System>
- <EventData>
  <Data>ffmpeg.exe</Data> 
  <Data>0.0.0.0</Data> 
  <Data>667ad11b</Data> 
  <Data>avfilter-9.dll</Data> 
  <Data>9.3.100.0</Data> 
  <Data>667ad11b</Data> 
  <Data>c0000005</Data> 
  <Data>000000000032dab4</Data> 
  <Data>3818</Data> 
  <Data>01dac77359173dbe</Data> 
  <Data>C:\Users\usr\Desktop\jellyfin-ffmpeg_6.0.1-7-portable_win64\ffmpeg.exe</Data> 
  <Data>C:\Users\usr\Desktop\jellyfin-ffmpeg_6.0.1-7-portable_win64\avfilter-9.dll</Data> 
  <Data>7b50b856-5989-4ef4-ad9f-03ded7842dfe</Data> 
  <Data /> 
  <Data /> 
  </EventData>
  </Event>
```

gnattu:
How long did you let it run? I cannot reproduce on my machine.

gnattu:
interesting, only occurs with windows builds

nyanmisaka:
It happens randomly.  Not sure if this has to do with gcc options and versions.

https://github.com/jellyfin/jellyfin-ffmpeg/blob/jellyfin/Dockerfile.win64.in#L20
https://github.com/jellyfin/jellyfin-ffmpeg/blob/jellyfin/builder/images/base-win64/Dockerfile#L43

Maybe try another builder? `./builder/makeimage.sh win64 gpl && ./builder/build.sh win64 gpl`

gnattu:
> It happens randomly.

The issue occurs frequently on Windows but never on Linux. Windows reports an access violation, which doesn't make sense to me if all memory accesses are to legal locations on Linux.

nyanmisaka:
> > It happens randomly.
> 
> The issue occurs frequently on Windows but never on Linux. Windows reports an access violation, which doesn't make sense to me if all memory accesses are to legal locations on Linux.

Does the compiler emit the same assembly code from the intrinsics?

gnattu:
`./builder/makeimage.sh win64 gpl` gives me `18.92 ../src/meson.build:28:12: ERROR: Program 'llvm-dlltool dlltool' not found or not executable` when building libplacebo...

nyanmisaka:
> `./builder/makeimage.sh win64 gpl` gives me `18.92 ../src/meson.build:28:12: ERROR: Program 'llvm-dlltool dlltool' not found or not executable` when building libplacebo...

I'll take a look at it later. You can drop the vulkan related scripts first.

gnattu:
The problem is now super stupid to me. I commented out both read and write to/from the framebuffer operation and it is still telling me access violation.

gnattu:
Well I found the cause. GCC generated code sequences  from `_mm256_extract_epi32` and `_mm_extract_epi32` intrinsics includes optimizations that do not play well with windows and will cause access violation. By storing the whole register to memory and read from there workarounds the access violation on windows.

nyanmisaka:
> Well I found the cause. GCC generated code sequences from `_mm256_extract_epi32` and `_mm_extract_epi32` intrinsics includes optimizations that do not play well with windows and will cause access violation. By storing the whole register to memory and read from there workarounds the access violation on windows.

I guess this uncertainty from the compiler is one of the reasons why upstream FFmpeg only accepts assembly code.

gnattu:
Well it turns out that it is not that simple.

Now I really believe it is due to gcc+windows. I can workaround this issue by reducing the `-filter_threads` to a lower value like 8. When this value is set to something like 16 or higher, it is super easy to trigger the access violation again.

To make it even worse, it seems like only cap the concurrency in tonemapx is not enough and this has to be a global option which means all filters in the chain has to be concurrency capped. 

With a global concurrency of 24 and we only spawn 1 job for tonemapx, the access violation still happens after a few moments.

nyanmisaka:
> Well it turns out that it is not that simple.
> 
> Now I really believe it is due to gcc+windows. I can workaround this issue by reducing the `-filter_threads` to a lower value like 8. When this value is set to something like 16 or higher, it is super easy to trigger the access violation again.
> 
> To make it even worse, it seems like only cap the concurrency in tonemapx is not enough and this has to be a global option which means all filters in the chain has to be concurrency capped.
> 
> With a global concurrency of 24 and we only spawn 1 job for tonemapx, the access violation still happens after a few moments.

Take it easy. We still have several months to investigate before JF 10.10.

Could it be related to LTO and GCC auto-vectorization?
https://github.com/jellyfin/jellyfin-ffmpeg/blob/jellyfin/debian/patches/0035-enable-gcc-vectorization-and-lto-auto.patch

gnattu:
`__attribute__((optimize("-fno-tree-vectorize")))` on simd functions made little to no difference

What works is that by implementing the write whole block back to memory and read from there(no usage of `_mm256_extract_epi32 `) plus a global threads cap of 8 and a normal video resolution (like 1920x1080), and an abnormal output of 1928x1080 requires even lower thread limit to 4.

I however observed that zscale can support a huge amount of concurrency (with 1920x1080 works with 24 and 1928x1080 works with 16) after the `_mm256_extract_epi32 ` workarounds being implemented. Maybe ffmpeg's native scaling filter is having some problem?

Edit: it seems like zscale works even without modification? At least it works with `-filter_threads 16` and all the resolution combination you mentioned.

nyanmisaka:
Have you seen these? MingGW-W64 seems to be quite fragile in handling AVX.

https://stackoverflow.com/questions/71859992/what-is-causing-this-memory-access-violation-error-0xc0000005-when-using-eigen

https://stackoverflow.com/questions/30928265/mingw64-is-incapable-of-32-byte-stack-alignment-required-for-avx-on-windows-x64

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54412

edit:
If this is the reason, a workaround is to add the compiler option `-Wa,-muse-unaligned-vector-move` to mingw.

https://github.com/search?q=use-unaligned-vector-move&type=code

gnattu:
> Have you seen these? MingGW-W64 seems to be quite fragile in handling AVX.
> 
> https://stackoverflow.com/questions/71859992/what-is-causing-this-memory-access-violation-error-0xc0000005-when-using-eigen
> 
> https://stackoverflow.com/questions/30928265/mingw64-is-incapable-of-32-byte-stack-alignment-required-for-avx-on-windows-x64
> 
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54412

Well it is worse than I think. The problem now is that the sse version also has such problems and I have no idea why. It does support higher concurrency though, but when the threads is too much the access violation will eventually come even with only sse. Still, zscale works better.

nyanmisaka:
> > Have you seen these? MingGW-W64 seems to be quite fragile in handling AVX.
> > https://stackoverflow.com/questions/71859992/what-is-causing-this-memory-access-violation-error-0xc0000005-when-using-eigen
> > https://stackoverflow.com/questions/30928265/mingw64-is-incapable-of-32-byte-stack-alignment-required-for-avx-on-windows-x64
> > https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54412
> 
> Well it is worse than I think. The problem now is that the sse version also has such problems and I have no idea why. It does support higher concurrency though, but when the threads is too much the access violation will eventually come even with only sse. Still, zscale works better.

Perhaps you should try compiling with MSVC to see if this is just another mingw gcc issue.
https://github.com/ShiftMediaProject/FFVS-Project-Generator

gnattu:
> Perhaps you should try compiling with MSVC to see if this is just another mingw gcc issue. https://github.com/ShiftMediaProject/FFVS-Project-Generator

Will figure it out later. `-Wa,-muse-unaligned-vector-move` does not help so this is another issue i think. But I still removed avx on windows due to its known bug.

gnattu:
The access violation with many threads even happens with msvc(btw our codebase is not really compatible with msvc and have to comment out a lot of things to make it compile). Perhaps I have to try that project generator to debug with visual studio to see what happens...

gnattu:
I "fixed" the access violation on Windows by refactoring the memory store logic and using more stable range clipping. 

The new memory store logic improved performance for all platforms:

i9-12900 with AVX2: 77fps->87fps (Linux)
M1 Max with NEON: 57fps->60fps

Now AVX is usable on Windows with compiler flag `-Wa,-muse-unaligned-vector-move`. The performance impact of this flag is around 5%. i9-12900 got 82fps on Windows.

gnattu:
A question about the yuv420p implementation: If most of the software decoder and encoders are not expecting p01x frames, isn't it safe to just drop the support for such frames and supports yuv420 exclusively? For what use case the p01x is preferred?

nyanmisaka:
> A question about the yuv420p implementation: If most of the software decoder and encoders are not expecting p01x frames, isn't it safe to just drop the support for such frames and supports yuv420 exclusively? For what use case the p01x is preferred?

It is safe. The original idea was to fall back to software filters for hybrid transcoding when the iGPU ran out of power but still had spare capacity for hardware decoding and encoding.

nyanmisaka:
> I "fixed" the access violation on Windows by refactoring the memory store logic and using more stable range clipping.
> 
> The new memory store logic improved performance for all platforms:
> 
> i9-12900 with AVX2: 77fps->87fps (Linux) M1 Max with NEON: 57fps->60fps
> 
> Now AVX is usable on Windows with compiler flag `-Wa,-muse-unaligned-vector-move`. The performance impact of this flag is around 5%. i9-12900 got 82fps on Windows.

That's incredible. Will applying this Wa globally affect the performance of other programs? It would be nice if we could add flags only for tonemapx.

gnattu:
> Will applying this Wa globally affect the performance of other programs?

This only affects ffmpeg itself and not other dependencies, and for ffmpeg itself, tonemapx is the only one that allows generating vectorized avx and most of the performance critical path is written with ASM in ffmpeg itself and using external libraries which does not have this applied. The only scenario that might be impacted by this is that the code is in ffmpeg and written in c, and gcc auto vectorized it using SSE2 and now the aligned load becomes unaligned. Even like this the perf impact should be negligible.

> It would be nice if we could add flags only for tonemapx.

It seems like there is no trivial way to do this. Do you know how could I do that?

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,3269 @@
...
++#include "internal.h"
++#include "video.h"
++
++#define REFERENCE_WHITE 203.0f
++#define FLOAT_EPS 1.175494351e-38f
++
++#if defined(__GNUC__) || defined(__clang__)
++#    if (__GNUC__ >= 11) || (__clang_major__ >= 12)
++#        define X86_64_V2 __attribute__((target("arch=x86-64-v2")))
++#        define X86_64_V3 __attribute__((target("arch=x86-64-v3")))
```
Bond-009: Still not a fan of targeting x96_64 feature levels while not checking all of the requirements at runtime

builder/variants/defaults-gpl.sh
```diff
@@ -6,3 +6,5 @@ GIT_BRANCH="jellyfin"
 LICENSE_FILE="COPYING.GPLv3"
 
 [[ $TARGET == linux* ]] && FF_CONFIGURE+=" --disable-libxcb --disable-xlib" || true
+[[ $TARGET == win* ]] && FF_CFLAGS+=" -Wa,-muse-unaligned-vector-move" || true
+[[ $TARGET == win* ]] && FF_CXXFLAGS+=" -Wa,-muse-unaligned-vector-move" || true
```
Bond-009: Maybe you should check for x86 too?

Bond-009 (dismissed):

builder/variants/defaults-gpl.sh
gnattu: We don't have windows arm build, gcc does not even support that

debian/patches/0080-add-tonemapx-filter.patch
gnattu: It is widely used. Microsoft MSVC does not even have the feature level, and if you enable AVX2 it will enable AVX2+FMA+BMI1+BMI2

gnattu:
Added direct support for yuv420p to eliminate the frame format conversion for software decoder/encoders. The performance of i9-12900 boosted to 95fps for 4K to 4K tonemapping on Linux. The M1 Max does not see a significant performance boost with this change (only around a 5% increase to 63fps), which is expected as it has three times the memory bandwidth of the i9. Therefore, reducing memory copy will not benefit it as much.

debian/patches/0080-add-tonemapx-filter.patch
gnattu: One valid use case for not using feature level is to support "fake" CPUs, like the CPUs in a VM. The hypervisor can disable instruction extensions as they wish to make them behave differently from all real CPUs. However, by doing so would break software compatibility not only for us.

builder/variants/defaults-gpl.sh
```diff
@@ -6,3 +6,5 @@ GIT_BRANCH="jellyfin"
 LICENSE_FILE="COPYING.GPLv3"
 
 [[ $TARGET == linux* ]] && FF_CONFIGURE+=" --disable-libxcb --disable-xlib" || true
+[[ $TARGET == win* ]] && FF_CFLAGS+=" -Wa,-muse-unaligned-vector-move" || true
```
nyanmisaka: https://github.com/jellyfin/jellyfin-ffmpeg/blob/4edf5f899ead175d22ee9ce8a56604365b940e92/docker-build-win64.sh#L591

The non ct-ng builder also requires Wa.
```
CFLAGS+=" -Wa,-muse-unaligned-vector-move"
CXXFLAGS+=" -Wa,-muse-unaligned-vector-move"

./configure ...
```
It seems that there is no easy way to add the flag only for specific files...

gnattu: Isn't this file also used by ct-ng?

If so we should add this flag exclusively to this file instead of defaults-gpl.sh

nyanmisaka:
> Added direct support for yuv420p to eliminate the frame format conversion for software decoder/encoders. The performance of i9-12900 boosted to 95fps for 4K to 4K tonemapping on Linux. The M1 Max does not see a significant performance boost with this change (only around a 5% increase to 63fps), which is expected as it has three times the memory bandwidth of the i9. Therefore, reducing memory copy will not benefit it as much.

The 12900 performs reasonably, while M1 Max seems to be bounded by its computing power/core counts.

Now LOC has expanded to 5 times of pure C impl. How about merging some duplicate lines?
For example, use macro to merge 10-to-10 and 10-to-8 funcs.
image

builder/variants/defaults-gpl.sh
nyanmisaka: Nope. ct-ng is scoped to `./builder`.

`docker-build-win64.sh` is a separate builder that uses Ubuntu's mingw instead of building the toolchain from scratch.

gnattu:
> The 12900 performs reasonably, while M1 Max seems to be bounded by its computing power/core counts.

Yes, in a VM with only 8P, the 12900 performs at about the same speed as the M1 Max

> How about merging some duplicate lines? For example, use macro to merge 10-to-10 and 10-to-8 funcs.

The problem with macros is that it hurts readability a lot and I really want to avoid using that, I know the upstream ffmpeg uses this a lot but their code is painful to read.

nyanmisaka:
> > The 12900 performs reasonably, while M1 Max seems to be bounded by its computing power/core counts.
> 
> 
> 
> Yes, in a VM with only 8P, the 12900 performs at about the same speed as the M1 Max
> 
> 
> 
> > How about merging some duplicate lines? For example, use macro to merge 10-to-10 and 10-to-8 funcs.
> 
> 
> 
> The problem with macros is that it hurts readability a lot and I really want to avoid using that, I know the upstream ffmpeg uses this a lot but their code is painful to read. 

So how about separating the different variants of simd code into multiple files `tonemapx_intrin_{sse,avx,neon}.c`? It's also painful to read them without a code editor (folding unrelated funcs).

gnattu:
> So how about separating the different variants of simd code into multiple files tonemapx_intrin_{sse,avx,neon}.c? It's also painful to read them without a code editor (folding unrelated funcs).

This sounds better to me

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++#include <float.h>
++#include <string.h>
++
++#include "libavutil/avassert.h"
++#include "libavutil/imgutils.h"
++#include "libavutil/internal.h"
++#include "libavutil/mem_internal.h"
++#include "libavutil/opt.h"
++#include "libavutil/cpu.h"
++#if  ARCH_AARCH64
```
nyanmisaka:
```suggestion
+#if ARCH_AARCH64
```

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++
++static float delinearize(float x, enum AVColorTransferCharacteristic trc_dst)
++{
++    if (trc_dst == AVCOL_TRC_BT709 || trc_dst == AVCOL_TRC_BT2020_10)
++        return inverse_eotf_bt1886(x);
++    else
++        return x;
++}
++
++static int comput_trc_luts(TonemapxContext *s, enum AVColorTransferCharacteristic trc_src,
```
nyanmisaka:
```suggestion
+static int compute_trc_luts(TonemapxContext *s, enum AVColorTransferCharacteristic trc_src,
```

a typo

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++        r_lin = MIX(r_lin, luma, coeff);
++        g_lin = MIX(g_lin, luma, coeff);
++        b_lin = MIX(b_lin, luma, coeff);avassert
++    }
++    float cmax = FFMAX(FFMAX(r_lin, g_lin), b_lin);
++    if (cmax > 1.0) {
++        r_lin /= cmax;
++        g_lin /= cmax;
++        b_lin /= cmax;
++    }*/
```
nyanmisaka:
```suggestion
```

Not used.

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++static const AVFilterPad tonemapx_outputs[] = {
++    {
++        .name         = "default",
++        .type         = AVMEDIA_TYPE_VIDEO,
++    },
++};
++
++AVFilter ff_vf_tonemapx = {
++    .name            = "tonemapx",
++    .description     = NULL_IF_CONFIG_SMALL("HDR to SDR tonemapping"),
```
nyanmisaka: Consider renaming the desc to reflect SIMD/intrinsics optimizations?

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
++ * Lesser General Public License for more details.
++ *
++ * You should have received a copy of the GNU Lesser General Public
++ * License along with FFmpeg; if not, write to the Free Software
++ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
++ */
++
++#ifndef FFMPEG_VF_TONEMAPX_INTRIN_NEON_H
++#define FFMPEG_VF_TONEMAPX_INTRIN_NEON_H
```
nyanmisaka: FFMPEG_VF => AVFILTER

`AVFILTER_TONEMAPX_INTRIN_NEON_H` is a common practice

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
+Index: FFmpeg/libavfilter/x86/Makefile
+===================================================================
+--- FFmpeg.orig/libavfilter/x86/Makefile
++++ FFmpeg/libavfilter/x86/Makefile
+@@ -39,6 +39,8 @@ OBJS-$(CONFIG_VOLUME_FILTER)
+ OBJS-$(CONFIG_V360_FILTER)                   += x86/vf_v360_init.o
+ OBJS-$(CONFIG_W3FDIF_FILTER)                 += x86/vf_w3fdif_init.o
+ OBJS-$(CONFIG_YADIF_FILTER)                  += x86/vf_yadif_init.o
++OBJS-$(CONFIG_TONEMAPX_FILTER)               += x86/vf_tonemapx_intrin_sse.o
++OBJS-$(CONFIG_TONEMAPX_FILTER)               += x86/vf_tonemapx_intrin_avx.o
```
nyanmisaka:
```
+OBJS-$(CONFIG_TONEMAPX_FILTER)               += x86/vf_tonemapx_intrin_sse.o \
                                                 x86/vf_tonemapx_intrin_avx.o
```

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++
++            robx4 = _mm_unpackhi_epi16(r0ox8, zero128);
++            gobx4 = _mm_unpackhi_epi16(g0ox8, zero128);
++            bobx4 = _mm_unpackhi_epi16(b0ox8, zero128);
++
++            yoax4 = _mm_mullo_epi32(roax4, _mm_set1_epi32(cry));
++            yoax4 = _mm_add_epi32(yoax4, _mm_mullo_epi32(goax4, _mm_set1_epi32(cgy)));
++            yoax4 = _mm_add_epi32(yoax4, _mm_mullo_epi32(boax4, _mm_set1_epi32(cby)));
++            yoax4 = _mm_add_epi32(yoax4, _mm_set1_epi32(out_rnd));
++            yoax4 = _mm_srai_epi32(yoax4, 21);
```
nyanmisaka: Use `out_sh` or hardcoded `21`?

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++            r0x4a = av_clip_int16_sse(r0x4a);
++
++            r1x4a = _mm_mullo_epi32(y1x4a, cyx4);
++            r1x4a = _mm_add_epi32(r1x4a, _mm_mullo_epi32(vx4a, _mm_set1_epi32(crv)));
++            r1x4a = _mm_add_epi32(r1x4a, rndx4);
++            r1x4a = _mm_srai_epi32(r1x4a, in_sh);
++            r1x4a = av_clip_int16_sse(r1x4a);
++
++            // g = av_clip_int16((y * cy + cgu * u + cgv * v + in_rnd) >> in_sh);
++            g0x4a = _mm_mullo_epi32(y0x4a, cyx4);
```
nyanmisaka: Duplicate `_mm_mullo_epi32()` ?
```
r0x4a = _mm_mullo_epi32(y0x4a, cyx4);
...
g0x4a = _mm_mullo_epi32(y0x4a, cyx4);
...
b0x4a = _mm_mullo_epi32(y0x4a, cyx4);
```

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
++ * Lesser General Public License for more details.
++ *
++ * You should have received a copy of the GNU Lesser General Public
++ * License along with FFmpeg; if not, write to the Free Software
++ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
++ */
++
++#ifndef FFMPEG_VF_TONEMAPX_H
++#define FFMPEG_VF_TONEMAPX_H
```
nyanmisaka: FFMPEG_VF => AVFILTER

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++
++#include "colorspace.h"
++
++#ifdef REFERENCE_WHITE
++#undef REFERENCE_WHITE
++#endif
++#define REFERENCE_WHITE 203.0f
```
nyanmisaka: IIRC the `REFERENCE_WHITE_ALT=203.0f` is only used by bt2390.
Other method such as hable and reinhard use `REFERENCE_WHITE=100.0f`.

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++
++#ifdef FLOAT_EPS
++#undef FLOAT_EPS
++#endif
++#define FLOAT_EPS 1.175494351e-38f
```
nyanmisaka: `1e-6f` in colorspace.h or `1.175494351e-38f`?

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++    int16_t (*yuv2rgb_coeffs)[3][3][8];
++    int16_t (*rgb2yuv_coeffs)[3][3][8];
++    double  (*rgb2rgb_coeffs)[3][3];
++    int rgb2rgb_passthrough;
++    const AVLumaCoefficients *coeffs, *ocoeffs;
++    double desat;
++} TonemapIntParams;
++
++void tonemap_frame_420p10_2_420p(uint8_t *dsty, uint8_t *dstu, uint8_t *dstv,
++                                        const uint16_t *srcy, const uint16_t *srcu, const uint16_t *srcv,
```
nyanmisaka:
```suggestion
+                                 const uint16_t *srcy, const uint16_t *srcu, const uint16_t *srcv,
```

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++            ux8 = _mm256_sub_epi32(ux8, in_uv_offx8);
++            vx8 = _mm256_sub_epi32(vx8, in_uv_offx8);
++
++            ux8a = _mm256_permutevar8x32_epi32(ux8, _mm256_set_epi32(3, 3, 2, 2, 1, 1, 0, 0));
++            ux8b = _mm256_permutevar8x32_epi32(ux8, _mm256_set_epi32(7, 7, 6, 6, 5, 5, 4, 4));
++            vx8a = _mm256_permutevar8x32_epi32(vx8, _mm256_set_epi32(3, 3, 2, 2, 1, 1, 0, 0));
++            vx8b = _mm256_permutevar8x32_epi32(vx8, _mm256_set_epi32(7, 7, 6, 6, 5, 5, 4, 4));
++
++            // r = av_clip_int16((y * cy + crv * v + in_rnd) >> in_sh);
++            r0x8a = _mm256_mullo_epi32(y0x8a, cyx8);
```
nyanmisaka:
```
r0x8a = _mm256_mullo_epi32(y0x8a, cyx8);
...
g0x8a = _mm256_mullo_epi32(y0x8a, cyx8);
...
b0x8a = _mm256_mullo_epi32(y0x8a, cyx8);
```
Not sure if the compiler will optimize this? There may be more like this in the patch.

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5601 @@
...
++    int crv = (*params->yuv2rgb_coeffs)[0][2][0];
++    int cgu = (*params->yuv2rgb_coeffs)[1][1][0];
++    int cgv = (*params->yuv2rgb_coeffs)[1][2][0];
++    int cbu = (*params->yuv2rgb_coeffs)[2][1][0];
++
++    int cry   = (*params->rgb2yuv_coeffs)[0][0][0];
++    int cgy   = (*params->rgb2yuv_coeffs)[0][1][0];
++    int cby   = (*params->rgb2yuv_coeffs)[0][2][0];
++    int cru   = (*params->rgb2yuv_coeffs)[1][0][0];
++    int ocgu  = (*params->rgb2yuv_coeffs)[1][1
```
nyanmisaka: FFMPEG_VF => AVFILTER

nyanmisaka:
There are also some warnings that can be eliminated before merging. There should be more in the GH actions log.

For example:
```
libavfilter/vf_tonemapx.c: In function ‘tonemap_frame_p016_p010_2_nv12’:
libavfilter/vf_tonemapx.c:495:13: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  495 |             int r00 = r[0], g00 = g[0], b00 = b[0];
      |             ^~~
libavfilter/vf_tonemapx.c: In function ‘tonemap_frame_420p10_2_420p’:
libavfilter/vf_tonemapx.c:585:13: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  585 |             int r00 = r[0], g00 = g[0], b00 = b[0];
      |             ^~~
libavfilter/vf_tonemapx.c: In function ‘tonemap_frame_420p10_2_420p10’:
libavfilter/vf_tonemapx.c:675:13: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  675 |             int r00 = r[0], g00 = g[0], b00 = b[0];
      |             ^~~
libavfilter/vf_tonemapx.c: In function ‘tonemap_frame_p016_p010_2_p016_p010’:
libavfilter/vf_tonemapx.c:767:13: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  767 |             int r00 = r[0], g00 = g[0], b00 = b[0];
      |             ^~~
libavfilter/vf_tonemapx.c: In function ‘filter_slice_planar10’:
libavfilter/vf_tonemapx.c:851:43: warning: passing argument 1 of ‘s->tonemap_func_planar10’ from incompatible pointer type [-Wincompatible-pointer-types]
  851 |     s->tonemap_func_planar10(out->data[0] + out->linesize[0] * slice_start,
      |                              ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                           |
      |                                           uint8_t * {aka unsigned char *}
libavfilter/vf_tonemapx.c:851:43: note: expected ‘uint16_t *’ {aka ‘short unsigned int *’} but argument is of type ‘uint8_t *’ {aka ‘unsigned char *’}
libavfilter/vf_tonemapx.c:852:43: warning: passing argument 2 of ‘s->tonemap_func_planar10’ from incompatible pointer type [-Wincompatible-pointer-types]
  852 |                              out->data[1] + out->linesize[1] * AV_CEIL_RSHIFT(slice_start, desc->log2_chroma_h),
libavfilter/vf_tonemapx.c:852:43: note: expected ‘uint16_t *’ {aka ‘short unsigned int *’} but argument is of type ‘uint8_t *’ {aka ‘unsigned char *’}
libavfilter/vf_tonemapx.c:853:43: warning: passing argument 3 of ‘s->tonemap_func_planar10’ from incompatible pointer type [-Wincompatible-pointer-types]
  853 |                              out->data[2] + out->linesize[2] * AV_CEIL_RSHIFT(slice_start, desc->log2_chroma_h),
libavfilter/vf_tonemapx.c:853:43: note: expected ‘uint16_t *’ {aka ‘short unsigned int *’} but argument is of type ‘uint8_t *’ {aka ‘unsigned char *’}
libavfilter/vf_tonemapx.c: In function ‘filter_slice_biplanar10’:
libavfilter/vf_tonemapx.c:870:45: warning: passing argument 1 of ‘s->tonemap_func_biplanar10’ from incompatible pointer type [-Wincompatible-pointer-types]
  870 |     s->tonemap_func_biplanar10(out->data[0] + out->linesize[0] * slice_start,
      |                                ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                             |
      |                                             uint8_t * {aka unsigned char *}
libavfilter/vf_tonemapx.c:870:45: note: expected ‘uint16_t *’ {aka ‘short unsigned int *’} but argument is of type ‘uint8_t *’ {aka ‘unsigned char *’}
libavfilter/vf_tonemapx.c:871:45: warning: passing argument 2 of ‘s->tonemap_func_biplanar10’ from incompatible pointer type [-Wincompatible-pointer-types]
  871 |                                out->data[1] + out->linesize[1] * AV_CEIL_RSHIFT(slice_start, desc->log2_chroma_h),
libavfilter/vf_tonemapx.c:871:45: note: expected ‘uint16_t *’ {aka ‘short unsigned int *’} but argument is of type ‘uint8_t *’ {aka ‘unsigned char *’}
```

debian/patches/0080-add-tonemapx-filter.patch
gnattu: I actually benchmarked merged and separated implementation and perf diff is not observable(<1fps) and therefore I don’t bother merging them in avx. I can merge these though.

debian/patches/0080-add-tonemapx-filter.patch
gnattu: 1.175494351e-38f is the one come with the tonemapx draft and that’s why I’m using this

debian/patches/0080-add-tonemapx-filter.patch
gnattu: The CPU implementation of hable and reinhard does not use reference white. It is doing it normalized with 1.0f. Should we use the REFERENCE_WHITE_ALT directly?

debian/patches/0080-add-tonemapx-filter.patch
nyanmisaka: `FLOAT_EPS` had not yet been added to colorspace.h when the draft was completed.
So let's use `1e-6f` to be consistent with other tonemap filters.

debian/patches/0080-add-tonemapx-filter.patch
gnattu: For arm neon, the right shift bit has to be a compile-time constant so it is hardcoded to 21 for all 8bit outputs. SSE and AVX does not have this constraint but I still prefer to make it a compile-time constant as the output depth is also a constant.

gnattu: If we want to reduce magic numbers, we can define it as macro in vf_tonemapx.h

debian/patches/0080-add-tonemapx-filter.patch
nyanmisaka: It seems that MPV now defaults to 203.0f, but FFmpeg is still using 100.0f. Using `REFERENCE_WHITE_ALT` looks fine to me.

debian/patches/0080-add-tonemapx-filter.patch
nyanmisaka: I saw those. Better add a comment near the hardcoded shift val to show where it comes from.

```
//    const int in_sh2 = 16 - in_depth; //<= Remove
...
            if (in_depth == 10) {
                // shift to low10bits for 10bit input
                // shift bit has to be compile-time constant
                y0x8 = vshrq_n_u16(y0x8, 6); // (16 - 10)
                y1x8 = vshrq_n_u16(y1x8, 6);
                uvx8 = vshrq_n_u16(uvx8, 6);
            }
```

Using macros is also acceptable, but it's a bit overkill.

nyanmisaka:
It seems that ffmpeg has a [`HAVE_INTRINSICS_NEON`](https://github.com/FFmpeg/FFmpeg/blob/2902ed25b58cb31ec3e0c2ef7d4740ae18200d2d/libavcodec/mpegvideo.c#L340) generated from [`configure`](https://github.com/FFmpeg/FFmpeg/blob/2902ed25b58cb31ec3e0c2ef7d4740ae18200d2d/configure#L6443). Is it necessary to add one for SSE and AVX as well? Just for completeness, in case our downstream users (e.g. SynoCommunity) are using legacy compilers and resulting in breakages.

edit:
- Disable corresponding SIMD code when header files or funcs are not available
- When disabling SIMD using ffmpeg flag, the corresponding code should be excluded `--disable-neon`
- Runtime SIMD check

gnattu:
> It seems that ffmpeg has a [`HAVE_INTRINSICS_NEON`](https://github.com/FFmpeg/FFmpeg/blob/2902ed25b58cb31ec3e0c2ef7d4740ae18200d2d/libavcodec/mpegvideo.c#L340) generated from [`configure`](https://github.com/FFmpeg/FFmpeg/blob/2902ed25b58cb31ec3e0c2ef7d4740ae18200d2d/configure#L6443). Is it necessary to add one for SSE and AVX as well? Just for completeness, in case our downstream users (e.g. SynoCommunity) are using legacy compilers and resulting in breakages.
> 
> edit:
> 
> * Disable corresponding SIMD code when header files or funcs are not available
> * When disabling SIMD using ffmpeg flag, the corresponding code should be excluded `--disable-neon`
> * Runtime SIMD check

I couldn't care less about those ancient compilers. I even dropped debian buster support due to its ancient gcc. Extend the macro checking for configure flag and disable with `--disable-neon` is probably all I'm going to do. Check for header existence is out of scope at least to me, and checking for function availability is almost impossible unless you write a super long configure script to test compile the whole file and then define the disable intrinsics macro, but I'm not going to do that

nyanmisaka:
> > It seems that ffmpeg has a [`HAVE_INTRINSICS_NEON`](https://github.com/FFmpeg/FFmpeg/blob/2902ed25b58cb31ec3e0c2ef7d4740ae18200d2d/libavcodec/mpegvideo.c#L340) generated from [`configure`](https://github.com/FFmpeg/FFmpeg/blob/2902ed25b58cb31ec3e0c2ef7d4740ae18200d2d/configure#L6443). Is it necessary to add one for SSE and AVX as well? Just for completeness, in case our downstream users (e.g. SynoCommunity) are using legacy compilers and resulting in breakages.
> > edit:
> > 
> > * Disable corresponding SIMD code when header files or funcs are not available
> > * When disabling SIMD using ffmpeg flag, the corresponding code should be excluded `--disable-neon`
> > * Runtime SIMD check
> 
> I couldn't care less about those ancient compilers. I even dropped debian buster support due to its ancient gcc. Extend the macro checking for configure flag and disable with `--disable-neon` is probably all I'm going to do. Check for header existence is out of scope at least to me, and checking for function availability is almost impossible unless you write a super long configure script to test compile the whole file and then define the disable intrinsics macro, but I'm not going to do that

Indeed, it is not practical to test all functions. How about checking with compiler version? When the gcc/clang version is lower than required, disable SIMD. Other compilers can be reasonably ignored.

gnattu:
> Indeed, it is not practical to test all functions. How about checking with compiler version? When the gcc/clang version is lower than required, disable SIMD. Other compilers can be reasonably ignored.

Done

docker-build-win64.sh
```diff
@@ -588,6 +588,9 @@ ffversion="$(dpkg-parsechangelog --show-field Version)"
 if [[ -f "patches/series" ]]; then
     quilt push -a
 fi
+# Workaround for GCC bug causing misaligned AVX instructions
+CFLAGS+=" -Wa,-muse-unaligned-vector-move"
+CXXFLAGS+=" -Wa,-muse-unaligned-vector-move"
```
nyanmisaka: How about we move this mingw gcc flag to https://github.com/FFmpeg/FFmpeg/blob/b818dff8d8e16f0831a379a5cdc15e813474a841/configure#L5795

```
# Workaround for GCC bug causing misaligned AVX instructions
enabled x86_64 && check_cflags -Wa,-muse-unaligned-vector-move
enabled x86_64 && check_cxxflags -Wa,-muse-unaligned-vector-move
```

debian/patches/0080-add-tonemapx-filter.patch
```diff
@@ -0,0 +1,5606 @@
...
++                                          int dstdepth, int srcdepth,
++                                          int width, int height,
++                                          const struct TonemapIntParams *params);
++
++void tonemap_frame_p016_p010_2_p016_p010(uint16_t *dsty, uint16_t *dstuv,
++                                                const uint16_t *srcy, const uint16_t *srcuv,
++                                                const int *dstlinesize, const int *srclinesize,
++                                                int dstdepth, int srcdepth,
++                                                int width, int height,
++                                                const struct TonemapIntParams *params);
```
nyanmisaka: missed indentation?
```suggestion
+void tonemap_frame_420p10_2_420p(uint8_t *dsty, uint8_t *dstu, uint8_t *dstv,
+                                 const uint16_t *srcy, const uint16_t *srcu, const uint16_t *srcv,
+                                 const int *dstlinesize, const int *srclinesize,
+                                 int dstdepth, int srcdepth,
+                                 int width, int height,
+                                 const struct TonemapIntParams *params);
+
+void tonemap_frame_p016_p010_2_nv12(uint8_t *dsty, uint8_t *dstuv,
+                                    const uint16_t *srcy, const uint16_t *srcuv,
+                                    const int *dstlinesize, const int *srclinesize,
+                                    int dstdepth, int srcdepth,
+                                    int width, int height,
+                                    const struct TonemapIntParams *params);
+
+void tonemap_frame_420p10_2_420p10(uint16_t *dsty, uint16_t *dstu, uint16_t *dstv,
+                                   const uint16_t *srcy, const uint16_t *srcu, const uint16_t *srcv,
+                                   const int *dstlinesize, const int *srclinesize,
+                                   int dstdepth, int srcdepth,
+                                   int width, int height,
+                                   const struct TonemapIntParams *params);
+
+void tonemap_frame_p016_p010_2_p016_p010(uint16_t *dsty, uint16_t *dstuv,
+                                         const uint16_t *srcy, const uint16_t *srcuv,
+                                         const int *dstlinesize, const int *srclinesize,
+                                         int dstdepth, int srcdepth,
+                                         int width, int height,
+                                         const struct TonemapIntParams *params);
```

docker-build-win64.sh
gnattu: I'd rather not touch the configure script. It is much more harder to maintain than our CI script

nyanmisaka: I agree. But we should also make the patch self-contained so that users can get reproducible results without using our build scripts, otherwise users will need to read our build scripts to realize the quirks exist.

debian/patches/0080-add-tonemapx-filter.patch
nyanmisaka: Does AVX2+FMA3 support imply x86-64-v3 in real hardware?
If not, does targeting `avx2,fma` instead of `x86-64-v3` hurt performance?

gnattu: > Does AVX2+FMA3 support imply x86-64-v3 in real hardware?

For all real CPUs from Intel and AMD, yes, all CPUs from them with both AVX2 and FMA3 conforms to the x86-64-v3 feature level. The level feature level is decided with the real CPUs after all.

gnattu:
I made a windows clang build for testing: https://github.com/gnattu/jellyfin-ffmpeg/releases/tag/win64-clang

This performs faster than the gcc version(at least on my own machine), but more testing is needed.

nyanmisaka:
[0080-add-simd-optimized-tonemapx-filter.patch](https://github.com/user-attachments/files/16071104/0080-add-simd-optimized-tonemapx-filter.patch)

I made some minor improvements:
- Briefly test `immtrin.h` availability instead of relying entirely on compiler version
- Comment out intrin code when disabling intrinsics, to prevent them from still being compiled and possibly throwing errors
- Fix `warning: unused variable 'cpu_flags'` when intrinsics is completely disabled

Two Q:
- Any chance to re-enable intrinsics for gcc 9? Our focal/20.04 build still uses it.
- Seems that `immintrin.h` already includes `emmintrin.h + smmintrin.h`, maybe just including `immintrin.h` is enough?

gnattu:
> Briefly test immtrin.h availability instead of relying entirely on compiler version

This seems unnecessary because you have to test compiler version after all. GCC has `immtrin.h` since 4.x (forget which precisely), but what is available in that header file varies largely which means intrinsics will be unavailable even with the presence of this file and you cannot test all instructions. Unless there is a stupid compiler that lies about itself and pretend to be a modern gcc/clang and does not implement x86 intrinsics. 

> Any chance to re-enable intrinsics for gcc 9? Our focal/20.04 build still uses it.

It *should* compile, though I cannot guarantee if it really works.

> Seems that immintrin.h already includes emmintrin.h + smmintrin.h, maybe just including immintrin.h is enough?

This should work on all modern compilers, not sure for ancient gcc

nyanmisaka:
> This seems unnecessary because you have to test compiler version after all. GCC has `immtrin.h` since 4.x (forget which precisely), but what is available in that header file varies largely which means intrinsics will be unavailable even with the presence of this file and you cannot test all instructions. Unless there is a stupid compiler that lies about itself and pretend to be a modern gcc/clang and does not implement x86 intrinsics.

At least this ensures that the current gcc/clang environment is capable of handling some kind of intrinsics before actually building ffmpeg, no?

If `immintrin.h` or one of its included headers is missing, or the package containing it is not installed or is corrupted, then simply checking the version of gcc/clang is not sufficient.

As for whether the contents of the headers are out of date, I think that is beyond our scope. If it's really necessary, we can add them manually, just like you did for `_mm_storeu_si32()` in gcc10-.

> It _should_ compile, though I cannot guarantee if it really works.

20.04 will reach EOL in April 2025. We can try enabling it and see.

> This should work on all modern compilers, not sure for ancient gcc

I think we don't care about those ancient compilers.

nyanmisaka (approved):

debian/patches/0080-add-tonemapx-filter.patch
nyanmisaka: @Bond-009 Do you have a consensus on this?

We actually use intrinsics up to SSE4.1 and AVX2+FMA. So conservatively speaking, if you don't want to use x86-64-v2/v3, we can use this instead:
```
__attribute__((target("sse4.2")))
__attribute__((target("avx2,fma")))
```
This is a subset of the former, but I haven't tested whether they have any impact on perf versus x86-64-v2/v3.

Bond-009: @nyanmisaka I'm totally fine with targeting the specific ISA extensions. I'm just not a fan of using x86_64 feature levels without checking all requirements. Especially if it's for improved performance but no benchmark results are provided.

nyanmisaka: > @nyanmisaka I'm totally fine with targeting the specific ISA extensions. I'm just not a fan of using x86_64 feature levels without checking all requirements. Especially if it's for improved performance but no benchmark results are provided.

@gnattu

gnattu: I did a benchmark and found no measurable perf difference between targeting ISA level and specific instructions. I pushed to always use instruction targets.

nyanmisaka (approved):

PR #408, opened 2024-06-28
Fix current CI

gnattu:
Our current CI is partially broken and this tries to fix it.

- Debian Buster is coming to EOL and we are not building server packages for it. The GCC it uses also lacks certain SIMD intrinsics required for the tonemapx filter.
- Harfbuzz dropped autoconf build in 9.0 and we are still using that.
- gnu website is not reliable

**Changes**

- Drop Debian Buster from CI build matrix.
- Use more reliable Linux Kernel mirror (mirrors.kernel.org) to replace the fragile ftp.gnu.org and ftpmirror.gnu.org
- Pin harfbuzz commit until we are ready to for its new version
- Use googlesource.com and github mirror instead of git.savannah.gnu.org

**Issues**

#407 Would require these changes to have the pipeline succeed.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,7 +22,7 @@ jobs:
     uses: ./.github/workflows/_meta.yaml
     with:
       distro: 'debian'
-      codenames: '["buster", "bullseye", "bookworm"]'
+      codenames: '["bullseye", "bookworm"]'
       architectures: '["amd64", "arm64", "armhf"]'
       release: false
 
--- a/docker-build-win64.sh
+++ b/docker-build-win64.sh
@@ -197,6 +197,7 @@ popd
 # HARFBUZZ
 git clone --depth=1 https://github.com/harfbuzz/harfbuzz.git
 pushd harfbuzz
+git checkout bc90b29b37fe3809f9e48aa7be08fbf2208e481a
 ./autogen.sh \
     --prefix=${FF_DEPS_PREFIX} \
     --host=${FF_TOOLCHAIN} \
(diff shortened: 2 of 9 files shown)


docker-build-win64.sh
```diff
@@ -195,8 +195,9 @@ popd
 popd
 
 # HARFBUZZ
-git clone --depth=1 https://github.com/harfbuzz/harfbuzz.git
+git clone https://github.com/harfbuzz/harfbuzz.git
```
nyanmisaka, line 199:
```suggestion
harfbuzz_commit="bc90b29b37fe3809f9e48aa7be08fbf2208e481a"
git clone https://github.com/harfbuzz/harfbuzz.git
```

This looks better to me.

docker-build-win64.sh
```diff
@@ -195,8 +195,9 @@ popd
...
 pushd harfbuzz
+git checkout bc90b29b37fe3809f9e48aa7be08fbf2208e481a
```
nyanmisaka:
```suggestion
git checkout ${harfbuzz_commit}
```

.github/workflows/build.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
     uses: ./.github/workflows/_meta.yaml
     with:
       distro: 'debian'
-      codenames: '["buster", "bullseye", "bookworm"]'
```
nyanmisaka, line 25: image

The oldoldstable/buster is reaching EOL. Also remove it from these scripts?
```
build
build.yaml
docker-build.sh
```

Shadowghost (approved):

nyanmisaka (approved):

PR #412, opened 2024-07-08
ci: Add Windows Clang builder

gnattu:
This adds a Windows native msys2 builder to CI which uses the clang64 environment and build the whole dependency chain with clang instead of gcc.

The dependencies are utilizing the makepkg build system and using the patches directly come from msys2. This way we can workaround known quirks for clang on windows easier and only care about to actually modify the build configs and pin to specific version we want.

Currently, most of the packages are pulled with versioned tarball instead of direct git commit pin.

This version includes all dependencies that the one built by our current docker builder has.

Current quirk of this version is that the `--enable-shared` option cannot be enabled for this as clang 's lld is not happy with the libfreetype being linked multiple times in a shared library, and we have to use static build all the way.

**Changes**

**Issues**


gnattu:
Currently we have to commit directly to the `.gitattribute` file due to the checkout maintainer's personal opinion to not add the option to force Unix LF Line ending: https://github.com/actions/checkout/issues/135. This does not fit our patch pushing workflow perfectly but this is probably fine because the upstream has not changed that file for 7 years.

nyanmisaka:
> Currently we have to commit directly to the `.gitattribute` file due to the checkout maintainer's personal opinion to not add the option to force Unix LF Line ending: [actions/checkout#135](https://github.com/actions/checkout/issues/135). This does not fit our patch pushing workflow perfectly but this is probably fine because the upstream has not changed that file for 7 years.

Does this workaround still work? https://github.com/actions/checkout/issues/135#issuecomment-602171132
```
    steps:
      - name: Set git to use LF
        run: |
          git config --global core.autocrlf false
          git config --global core.eol lf

      - uses: actions/checkout@v2
```

gnattu:
> Does this workaround still work? [actions/checkout#135 (comment)](https://github.com/actions/checkout/issues/135#issuecomment-602171132)
> 
> ```
>     steps:
>       - name: Set git to use LF
>         run: |
>           git config --global core.autocrlf false
>           git config --global core.eol lf
> 
>       - uses: actions/checkout@v2
> ```

No. I've tried it with [multiple variants](https://github.com/jellyfin/jellyfin-ffmpeg/pull/412/commits/46bb615836fc8be716e889e73b0a02f2be84e92e) and it insists to checkout as crlf, and the maintainer of the checkout CI action has strong preference on directly committing  it into the `.gitattribute` and is unlikely to fix this workaround if it does not work.

Shadowghost (approved):
LGTM as long as it works 😄

nyanmisaka:
The libass/vf_subtitles seems unhappy: (font provider **fontconfig**)
```
[Parsed_subtitles_0 @ 0000017bddca6380] libass API version: 0x1703000
[Parsed_subtitles_0 @ 0000017bddca6380] libass source: tarball: 0.17.3
[Parsed_subtitles_0 @ 0000017bddca6380] Shaper: FriBidi 1.0.15 (SIMPLE) HarfBuzz-ng 9.0.0 (COMPLEX)
Fontconfig error: Cannot load default config file: No such file: (null)
[Parsed_subtitles_0 @ 0000017bddca6380] No usable fontconfig configuration file found, using fallback.
Fontconfig error: Cannot load default config file: No such file: (null)
[Parsed_subtitles_0 @ 0000017bddca6380] Using font provider fontconfig
```

mingw build: (font provider **directwrite (with GDI)**)
```
[Parsed_subtitles_0 @ 0000017c4edede80] libass API version: 0x1703000
[Parsed_subtitles_0 @ 0000017c4edede80] libass source: commit: 0.17.3-0-ge46aedea0a0d17da4c4ef49d84b94a7994664ab5
[Parsed_subtitles_0 @ 0000017c4edede80] Shaper: FriBidi 1.0.15 (SIMPLE) HarfBuzz-ng 8.5.0 (COMPLEX)
[Parsed_subtitles_0 @ 0000017c4edede80] Using font provider directwrite (with GDI)
```

gnattu:
> mingw build: (font provider directwrite (with GDI))

I enabled directwrite and libass is working now.

nyanmisaka (approved):
YOLO

PR #424, opened 2024-08-01
New upstream version 7.0.2

### LINKED ISSUE #391: Upgrade to FFmpeg 7.0
Hi! I'm wondering what the current thoughts are on updating FFmpeg to 7.0? Following #306, the conclusion was to skip 6.1 and upgrade directly to 7.0. Is there anything blocking this upgrade?

For context, I ask because it has many improvements for Vulkan that I'd love to take advantage of (the long-term goal being to use Vulkan for decoding and filters across all devices). I can't use a normal FFmpeg build since my project relies on the RKMPP encoder here and possibly some of the other tweaks for hardware acceleration.

nyanmisaka:
Vulkan patches are not yet ready.

**Changes**
- New upstream version 7.0.2
- Rename package to jellyin-ffmpeg7

**Issues**
- Fixes #391


builder/scripts.d/45-harfbuzz.sh
```diff
@@ -29,7 +29,12 @@ ffbuild_dockerbuild() {
         -Dutilities=disabled
     )
 
-    if [[ $TARGET == win* || $TARGET == linux* ]]; then
+    if [[ $TARGET == win* ]]; then
+        myconf+=(
+            --cross-file=/cross.meson
+            -Dgdi=enabled
```
gnattu: I don't think enable gdi is correct because it does not do what we want to per harfbuzz doc: https://harfbuzz.github.io/integration-uniscribe.html

This option is only for test purposes and is not the Uniscribe-integration function and is only used in the `hb-shape` cli tool.

builder/scripts.d/50-libbluray.sh
```diff
@@ -11,6 +11,8 @@ ffbuild_dockerbuild() {
     git-mini-clone "$SCRIPT_REPO" "$SCRIPT_COMMIT" libbluray
     cd libbluray
 
+    sed -i 's/dec_init/libbluray_dec_init/g' src/libbluray/disc/*.c src/libbluray/disc/*.h
```
gnattu: macOS need to specify gsed as this won't work with BSD sed

builder/images/base-win64/Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM $GH_REPO/base:latest
 
 RUN --mount=src=ct-ng-config,dst=/.config \
     git clone --filter=blob:none https://github.com/crosstool-ng/crosstool-ng.git /ct-ng && cd /ct-ng && \
+    curl "https://gcc.gnu.org/git/?p=gcc.git;a=patch;h=747c4b58573ea00419f64293a61537eb69f43307" > packages/gcc/14.1.0/9999-fix-inline.patch && \
```
gnattu: Let's hope gnu.org won't down again due to this. Can we just self-contain this patch?

debian/patches/0051-add-coreimage-based-vf-transpose-vt-filter.patch
```diff
@@ -0,0 +1,573 @@
...
+ transpose_opencl_filter_deps="opencl"
+ transpose_vaapi_filter_deps="vaapi VAProcPipelineCaps_rotation_flags"
+-transpose_vt_filter_deps="videotoolbox VTPixelRotationSessionCreate"
++transpose_vt_filter_deps="coreimage videotoolbox"
+ transpose_vulkan_filter_deps="vulkan spirv_compiler"
+ unsharp_opencl_filter_deps="opencl"
+ uspp_filter_deps="gpl avcodec"
+Index: FFmpeg/libavfilter/vf_transpose_vt.c
+===================================================================
+--- FFmpeg.orig/libavfilter/vf_transpose_vt.c
```
gnattu, line 16: I'm going to make a version merges two of them. But this is good enough for now.

builder/images/base-win64/Dockerfile
nyanmisaka: let's use the gh mirror https://github.com/gcc-mirror/gcc/commit/747c4b58573ea00419f64293a61537eb69f43307

gnattu:
With 7.0.1+backported patches ffmpeg now correctly set the buffer attachments for videotoolbox and now the tonemap pipeline needs to be slightly modified because the old one works with some assumptions with wrong colors, and the HDR attachments is not properly removed doing tone-mapping to 10bit now.

nyanmisaka:
> and the HDR attachments is not properly removed doing tone-mapping to 10bit now.

Is `CVBufferRemoveAttachment()` needed [here](https://github.com/nyanmisaka/jellyfin-ffmpeg/blob/d103afea1dc01ec1abf18fb6bfed9e5170547d10/debian/patches/0052-add-vf-tonemap-videotoolbox-filter.patch#L1790-L1793)?

gnattu:
> > and the HDR attachments is not properly removed doing tone-mapping to 10bit now.
> 
> Is `CVBufferRemoveAttachment()` needed [here](https://github.com/nyanmisaka/jellyfin-ffmpeg/blob/d103afea1dc01ec1abf18fb6bfed9e5170547d10/debian/patches/0052-add-vf-tonemap-videotoolbox-filter.patch#L1790-L1793)?

We need to override the `kCVImageBufferTransferFunctionKey` I think, but I will look at this later. Once this one is merged  I will submit a PR fix it.

gnattu:
Well interesting, `scale_vt` is still producing wrong color without manually hwupload. I guess the upstream missed up something once again.

nyanmisaka:
> Well interesting, `scale_vt` is still producing wrong color without manually hwupload. I guess the upstream missed up something once again.

These two fixes have been included in the backport patches.
https://github.com/FFmpeg/FFmpeg/commit/1fa7554bd6cc35b008045d33f9e948a92e7b901c
https://github.com/FFmpeg/FFmpeg/commit/cd9ceaef22ecc25278c771169d179dbfdb24a355

Is this possible related to the newly added SDK version check?
https://github.com/FFmpeg/FFmpeg/commit/85706f5136cf7c88f95843b2634dd3f7d7d2cb6d
https://github.com/FFmpeg/FFmpeg/commit/2fc37c42391a38bfe0aea248f2fe75c9033f98e5
https://github.com/FFmpeg/FFmpeg/commit/ca7fcf50891997eb95b4bf6d5cd4e2446e8c8631

gnattu:
> > Well interesting, `scale_vt` is still producing wrong color without manually hwupload. I guess the upstream missed up something once again.
> 
> These two fixes have been included in the backport patches. [FFmpeg/FFmpeg@1fa7554](https://github.com/FFmpeg/FFmpeg/commit/1fa7554bd6cc35b008045d33f9e948a92e7b901c) [FFmpeg/FFmpeg@cd9ceae](https://github.com/FFmpeg/FFmpeg/commit/cd9ceaef22ecc25278c771169d179dbfdb24a355)
> 
> Is this possible related to the newly added SDK version check? [FFmpeg/FFmpeg@85706f5](https://github.com/FFmpeg/FFmpeg/commit/85706f5136cf7c88f95843b2634dd3f7d7d2cb6d) [FFmpeg/FFmpeg@2fc37c4](https://github.com/FFmpeg/FFmpeg/commit/2fc37c42391a38bfe0aea248f2fe75c9033f98e5) [FFmpeg/FFmpeg@ca7fcf5](https://github.com/FFmpeg/FFmpeg/commit/ca7fcf50891997eb95b4bf6d5cd4e2446e8c8631)

The CGColorSpace generated from its provided buffer is wrong and creates an SDR colorspace.

nyanmisaka:
> The CGColorSpace generated from its provided buffer is wrong and creates an SDR colorspace.

Is this also reproducible in the ffmpeg master branch? I saw your discussion on the mailing list and thought it was fixed.

gnattu:
> > The CGColorSpace generated from its provided buffer is wrong and creates an SDR colorspace.
> 
> Is this also reproducible in the ffmpeg master branch? I saw your discussion on the mailing list and thought it was fixed.

I identified the problem but I found that it is really funny. The whole sdk version checking drama is to get the attachments from the pixel buffer, but the attachments get from that buffer will always creates a non-HDR  buffer. The very first implementation allocated another dedicated dictionary which only specifies the color primaries, the transfer function and the color matrix, which will properly create an HDR colorspace and make scale_vt happy. I need to submit this to the upstream I think.

gnattu:
The good news is that, a framebuffer with properly set attachments brings slightly performance improvements over hwupload, instead of regression in earlier version. So maybe we can get rid of hwupload in jellyfin 10.10.

gnattu:
Patch submitted to upstream: https://ffmpeg.org//pipermail/ffmpeg-devel/2024-August/331960.html

nyanmisaka:
> Patch submitted to upstream: https://ffmpeg.org//pipermail/ffmpeg-devel/2024-August/331960.html

defined but not used?
```c
static CFDictionaryRef vt_cv_buffer_copy_attachments(CVBufferRef buffer,
                                                     CVAttachmentMode attachment_mode)
```

redundant braces?
```c
#if (TARGET_OS_OSX && __MAC_OS_X_VERSION_MAX_ALLOWED >= 100800) || \
    (TARGET_OS_IOS && __IPHONE_OS_VERSION_MAX_ALLOWED >= 100000)
    if (__builtin_available(macOS 10.8, iOS 10, *)) {
        colorspace = CVImageBufferCreateColorSpaceFromAttachments(attachments);
    }
#endif
```

gnattu:
> > Patch submitted to upstream: https://ffmpeg.org//pipermail/ffmpeg-devel/2024-August/331960.html
> 
> defined but not used?
> 
> ```c
> static CFDictionaryRef vt_cv_buffer_copy_attachments(CVBufferRef buffer,
>                                                      CVAttachmentMode attachment_mode)
> ```
> 

This can be used because the real cause is simpler than I originally thought.

> redundant braces?
> 
> ```c
> #if (TARGET_OS_OSX && __MAC_OS_X_VERSION_MAX_ALLOWED >= 100800) || \
>     (TARGET_OS_IOS && __IPHONE_OS_VERSION_MAX_ALLOWED >= 100000)
>     if (__builtin_available(macOS 10.8, iOS 10, *)) {
>         colorspace = CVImageBufferCreateColorSpaceFromAttachments(attachments);
>     }
> #endif
> ```

`CVImageBufferCreateColorSpaceFromAttachments` is only available in macOS 10.8 so this is still necessary.

gnattu:
See: https://github.com/FFmpeg/FFmpeg/blob/1b2a925e94c772c59a88c03c1654bddf6aff0ca2/libavutil/hwcontext_videotoolbox.c#L600

The color primaries value is wrongly assigned to color transfer key and that's why the attachments copied from the pixel buffer will always create an SDR colorspace. After correcting this the original code works properly.

nyanmisaka:
> See: https://github.com/FFmpeg/FFmpeg/blob/1b2a925e94c772c59a88c03c1654bddf6aff0ca2/libavutil/hwcontext_videotoolbox.c#L600
> 
> 
> 
> The color primaries value is wrongly assigned to color transfer key and that's why the attachments copied from the pixel buffer will always create an SDR colorspace. After correcting this the original code works properly.

`colorxxx` is hard to distinguish at a glance... It's necessary to use underscores or camelCase.

gnattu:
> Patch submitted to upstream: https://ffmpeg.org//pipermail/ffmpeg-devel/2024-August/331960.html

Replaced by: https://ffmpeg.org//pipermail/ffmpeg-devel/2024-August/331962.html

gnattu:
This version currently does not contain the nop `vt_device_derive` because it is not needed for upstream. But our current server code uses `hwupload=derive_device=videotoolbox` a lot. Should we add that as well to support current server?

nyanmisaka:
> This version currently does not contain the nop `vt_device_derive` because it is not needed for upstream. But our current server code uses `hwupload=derive_device=videotoolbox` a lot. Should we add that as well to support current server?

I remembered it. Will re-add it as a separate patch since upstream will never accept it.

nyanmisaka:
Merging to `jellyfin-7.0` branch for now.

PR #430, opened 2024-08-08
avfilter/tonemap*: add ICtCp based tonemap method

### LINKED ISSUE #429: Opportunity for a new default tonemap mode
Jellyfin currently has two tonemap modes: MAX mode, which applies TMO to the maximum of the RGB channels, and RGB mode, which applies TMO to each RGB channel individually. However, both methods have significant shortcomings in certain cases. A new mode that applies TMO on the relative luminance has recently been merged in #418 and #416. This mode resolves the distorted color/overly saturated color issue in RGB mode and the significant hue shift issue on bright pixels in MAX mode. 

My current working draft introduces a new mode called ITP mode. This mode does not operate in the linear RGB color space like the other methods but operates in the ICtCp color space. The TMO is applied to the Intensity of the ICtCp color. This mode also has a special desaturation method that preserves hue in ICtCp space, which should be closer to human perception due to the design of this color space. This tonemap mode preserves colors in bright pixels even better than the LUM mode, and I propose to make this mode the new default for Jellyfin 10.10 and jellyfin-ffmpeg 7.0. 

The major drawback of this mode is the extra computation required for the color space conversion. However, for BT.2390 methods, this is not much slower because this color space does not need the extra non-linearity applied to the signal like other modes, as such non-linearity functions are already applied during color space conversion. For other TMOs that expect a linear signal, extra linearization and de-linearization are required, which makes this mode slower. For users with lower spec GPUs, I suggest them to use the LUM mode which is a good enough alternative for common cases.

To demonstrate the effectiveness of this tonemap mode, I made some sample scene images as examples.

Let's begin with the customizable desaturation parameter. Look at the following furnace example with `desat=0.5`:

**RGB mode:**
desat_rgb

**MAX mode:**
desat_max

**LUM mode:**
desat_lum

**ITP mode:**
desat_itp

As we can see from this example, ICtCp-based desaturation shows a significant advantage over RGB-based desaturation. LUM and MAX modes have a significant hue shift, causing the furnace to appear too red. The RGB mode does not experience as severe a hue shift, but the desaturation is not effective as the color remains overly saturated. ITP mode preserves the hue while gradually turning the overly bright furnace color to white, producing most naturally looking color.

In fact, for this scene, the ICtCp-based tone mapping does not have over saturation problem in the first place, I made this sample mostly to show the hue preserving capability of ICtCp. This is the results of `desat=0`:

**RGB mode:**
origin_rgb

**MAX mode:**
origin_max

**LUM mode:**
origin_lum

**ITP mode:**
origin_itp

As you can see, the ITP mode produces natural-looking colors and does not require extra desaturation.

-----

To demonstrate the detail-preserving capability for bright pixels, let look at this glass blowing scene:

**RGB mode:**
rgb

**MAX mode:**
max

**LUM mode:**
lum

**ITP mode:**
itp

As you can see, RGB mode preserves highlights but produces unnatural colors. MAX mode loses most of the highlights, and while LUM mode preserves highlights better than MAX, it is still not as effective as ITP mode.

Another example is the campfire scene:

**RGB mode:**
rgb

**MAX mode:**
max

**LUM mode:**
lum

**ITP mode:**
itp

Pay attention to the details of the flame in the bottom right corner.

-------

To further demonstrate the limitation of current modes (MAX and RGB), lets look at this scene that direct captures the sun:

**RGB mode:**
rgb

**MAX mode:**
max

**LUM mode:**
lum

**ITP mode:**
itp

As you can see, both of the current modes render sunlight poorly. Both LUM and ITP modes improve the sunlight significantly. In this case, LUM mode has a darker sun border, which appears less "natural" but is not subjectively bad.

Direct capture of the sun is an extreme case, but the current modes have more limitations. Let's look at this bridge scene shot at night:

**RGB mode:**
rgb

**MAX mode:**
max

**LUM mode:**
lum

**ITP mode:**
itp

RGB mode is too dark for a night scene, and MAX mode renders the lights on the road way too red. Both LUM and ITP modes produce much more natural colors in this case. For this scene, the LUM and ITP results are not visually distinguishable.

For more scene comparisons, I have made a zip file that includes additional scenes not shown here:[tonemode_comparison_extra.zip](https://github.com/user-attachments/files/16516217/tonemode_comparison_extra.zip)

If there is no objection I will make the ITP mode as the default tonemap mode for jellyfin-ffmpeg7 and Jellyfin 10.10.

gnattu:
This adds ICtCp based tonemap to OpenCL, Metal and CUDA based tonemap filters, and will use the mode as the new default. On GPUs with lower specs, OpenCL and CUDA implementation will default to relative luminance mode. The Metal implementation does not have a performance tradeoff mode and will always default to ICtCp mode.

This also fixed attachment related issue on Metal when output is 10bit by always setting the attachement values.

The ICtCp mode's desat algorithm is different from other modes. The coefficient applied to the chroma is defined by a Gaussian Function in the form:

```math
f(I) = \exp\left(\frac{-\left(I-\frac{L_{\text{target}} - p_{\text{desat}}}{2}\right)^2}{2L_{\text{peak}}}\right)
```

Where $I$ is the ICtCp intensity linearized by ST2084 EOTF, $p_{\text{desat}}$ is the desat parameter, $L_{\text{target}}$ is the target peak luminance (1.0 for SDR) and $L_{\text{peak}}$ is the HDR peak luminance for current frame. 

The current curve is tuned to preserve more color detail to preserve natural looking than to aggressively turning bright colors into white, and is different from the initial curve  implemented in #429. 

**Original**
image

**`desat=0.5` in #429**
image

**`desat=0.5` current**
desat_itp_g2

This curve is not a standardized practice and is purely my own creation, which means it could lead to unpleasant results for certain cases I've not tested on.

**Changes**

**Issues**

Fixes #429

### DIFF
--- a/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
+++ b/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
@@ -4,13 +4,13 @@ Index: FFmpeg/libavfilter/opencl.c
 +++ FFmpeg/libavfilter/opencl.c
 @@ -169,7 +169,7 @@ int ff_opencl_filter_load_program(AVFilt
      }
- 
+
      cle = clBuildProgram(ctx->program, 1, &ctx->hwctx->device_id,
 -                         NULL, NULL, NULL);
 +                         "-cl-finite-math-only -cl-unsafe-math-optimizations", NULL, NULL);
      if (cle != CL_SUCCESS) {
          av_log(avctx, AV_LOG_ERROR, "Failed to build program: %d.\n", cle);
- 
+
 @@ -330,7 +330,7 @@ void ff_opencl_print_const_matrix_3x3(AV
      av_bprintf(buf, "__constant float %s[9] = {\n", name_str);
      for (i = 0; i < 3; i++) {
@@ -26,7 +26,7 @@ Index: FFmpeg/libavfilter/opencl.h
 +++ FFmpeg/libavfilter/opencl.h
 @@ -206,17 +206,17 @@ do {
  } while(0)
- 
+
  /**
 - * Perform a blocking write to a buffer.
 + * Perform a blocking write to a buffer with offset.
@@ -47,7 +47,7 @@ Index: FFmpeg/libavfilter/opencl.h
          0,                                                                                      \
 @@ -227,6 +227,15 @@ do {
  } while(0)
- 
+
  /**
 + * Perform a blocking write to a buffer.
 + *
@@ -67,7 +67,7 @@ Index: FFmpeg/libavfilter/opencl/colorspace_common.cl
 +++ FFmpeg/libavfilter/opencl/colorspace_common.cl
 @@ -17,7 +17,17 @@
   */
- 
+
  #define ST2084_MAX_LUMINANCE 10000.0f
 -#define REFERENCE_WHITE 100.0f
 +#define ST2084_M1 0.1593017578125f
@@ -81,13 +81,13 @@ Index: FFmpeg/libavfilter/opencl/colorspace_common.cl
 +#define ARIB_B67_C 0.55991073f
 +
 +#define FLOAT_EPS 1e-6f
- 
+
  #if chroma_loc == 1
      #define chroma_sample(a,b,c,d) (((a) + (c)) * 0.5f)
-@@ -33,80 +43,106 @@
+@@ -33,81 +43,124 @@
      #define chroma_sample(a,b,c,d) (((a) + (b) + (c) + (d)) * 0.25f)
  #endif
- 
+
 -constant const float ST2084_M1 = 0.1593017578125f;
 -constant const float ST2084_M2 = 78.84375f;
 -constant const float ST2084_C1 = 0.8359375f;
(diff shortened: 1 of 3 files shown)


Shadowghost (approved):

debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
```diff
@@ -220,59 +249,46 @@ Index: FFmpeg/libavfilter/opencl/colorspace_common.cl
...
 +    float r = linearize(c.x);
 +    float g = linearize(c.y);
 +    float b = linearize(c.z);
-+    return lrgb2lrgb((float3)(r, g, b));
-+#else
-+    return lrgb2lrgb((float3)(c.x, c.y, c.z));
++    return (float3)(r, g, b);
+ #else
+-    return c;
++    return (float3)(c.x, c.y, c.z);
```
```suggestion
+    return c;
```

debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
```diff
@@ -570,33 +630,76 @@ Index: FFmpeg/libavfilter/opencl/tonemap.cl
...
+     }
+-    return r;
++#ifdef TONE_FUNC_BT2390
++        float src_peak_delin_pq = inverse_eotf_st2084(peak);
++        float dst_peak_delin_pq = inverse_eotf_st2084(1.0f);
++        MAP_FOUR_PIXELS(i4, src_peak_delin_pq, dst_peak_delin_pq)
++#else
++        i4 = eotf_st2084x4(i4);
++        MAP_FOUR_PIXELS(i4, peak, 1.0f)
++        i4 = inverse_eotf_st2084x4(i4);
```
```suggestion
+    float src_peak_delin_pq = inverse_eotf_st2084(peak);
+    float dst_peak_delin_pq = inverse_eotf_st2084(1.0f);
+    MAP_FOUR_PIXELS(i4, src_peak_delin_pq, dst_peak_delin_pq)
+#else
+    i4 = eotf_st2084x4(i4);
+    MAP_FOUR_PIXELS(i4, peak, 1.0f)
+    i4 = inverse_eotf_st2084x4(i4);
```

nyanmisaka (approved):

PR #431, opened 2024-08-08
avfilter/transpose_vt: select implementation based on macOS version

gnattu:
**Changes**

This combines the VT native rotation and CoreImage based rotation implementation and will seletect based on the running macOS version. On macOS 13+ this will choose the faster VT native one and on macOS 12 it will choose the CoreImage one. CI is also modified for this change.

**Issues**


gnattu:
Targets master for CI testing, will re-target after CI finished.

nyanmisaka (approved):

PR #439, opened 2024-08-15
avfilter/*videotoolbox: fix color range handling

gnattu:
**Changes**

FFmpeg's implementation of the VideoToolbox hardware context requires specifying the color range during the context's initialization, rather than deriving it from the `AVFrame`. Unfortunately, both the upstream-provided VideoToolbox filters and our custom filters did not correctly handle the color range when creating the VT context, leaving it unspecified. As a result, the generated pixels default to the TV range, since that is the default value.

This PR makes most of the filters to inherit the input's color range, and makes the `tonemap_videotoolbox` filter to correctly set hardware color range based on user input. It was previously defaults to tv range, not it defaults to be same as input.

This PR also fixed the double-free issue in the upstream `yadif_videotoolbox` filter. That filer frees the auto-released objc objects which may cause segfaults.

I plan to submit the fix for upstream filters to upstream mailing list shortly as well. If we are going to make more 6.x releases, I will create another PR to back-port these changes to our 6.x branch. 

**Issues**


gnattu:
This leaves unnoticed for so long is probably most videos are in the tv range and the full range ones are super rare.

gnattu:
Upstream patches sent: https://ffmpeg.org//pipermail/ffmpeg-devel/2024-August/332373.html

Shadowghost (approved):

nyanmisaka:
https://ffmpeg.org/pipermail/ffmpeg-devel/2024-August/332375.html
Does this review need to be resolved?

gnattu:
> https://ffmpeg.org/pipermail/ffmpeg-devel/2024-August/332375.html Does this review need to be resolved?

This file is not related to this PR though. The upstream want another method to resolved the leak because they want "future use" but it is okay for us now. I will submit another version when I have time.

debian/patches/0052-add-vf-tonemap-videotoolbox-filter.patch
```diff
@@ -2040,8 +2046,8 @@ Index: FFmpeg/libavfilter/vf_tonemap_videotoolbox.m
 +    { "p",         "Set color primaries", OFFSET(primaries), AV_OPT_TYPE_INT, { .i64 = AVCOL_PRI_BT709 }, -1, INT_MAX, FLAGS, .unit = "primaries" },
 +        { "bt709",            0,       0,                 AV_OPT_TYPE_CONST, { .i64 = AVCOL_PRI_BT709 },         0, 0, FLAGS, .unit = "primaries" },
 +        { "bt2020",           0,       0,                 AV_OPT_TYPE_CONST, { .i64 = AVCOL_PRI_BT2020 },        0, 0, FLAGS, .unit = "primaries" },
-+    { "range",         "Set color range", OFFSET(range), AV_OPT_TYPE_INT, { .i64 = AVCOL_RANGE_MPEG }, -1, INT_MAX, FLAGS, .unit = "range" },
-+    { "r",             "Set color range", OFFSET(range), AV_OPT_TYPE_INT, { .i64 = AVCOL_RANGE_MPEG }, -1, INT_MAX, FLAGS, .unit = "range" },
++    { "range",         "Set color range", OFFSET(range), AV_OPT_TYPE_INT, { .i64 = -1 }, -1, INT_MAX, FLAGS, .unit = "range" },
```
nyanmisaka, line 2049: Maybe also make cuda and ocl tonemap follow this behavior?

nyanmisaka (approved):

debian/patches/0052-add-vf-tonemap-videotoolbox-filter.patch
gnattu: done

nyanmisaka (approved):

PR #447, opened 2024-09-03
avfilter/vf_tonemapx: add dovi tonemapx for 420p10 inputs

gnattu:
**Changes**

This adds Dolby Vision reshaping support for DV Profile 5 videos. The implementation includes some trade-offs for performance during color conversion calculations, which may cause colors in very bright or very dark scenes to appear off. The current implementation requires the input to be in 420p10 format to perform the reshaping. Expanding support to p010 inputs is possible if a use case arises.

Unlike the GPGPU implementation, which performs BL reshaping when only BL is required, this implementation only performs BL reshaping when absolutely necessary. For example, it will not reshape BL for Profile 8 videos and will instead use the HDR10 fallback for performance reasons. The additional reshaping incurs a performance loss of about 35%-50% compared to HDR10 inputs that do not require reshaping. Users will need a 16-core Zen3-level CPU to achieve 4K60 playback if they want to tonemap DV P5 videos on CPU.

**Issues**

Closes #404

### DIFF
--- a/debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
+++ b/debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
@@ -95,7 +95,7 @@ Index: FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
-@@ -0,0 +1,2150 @@
+@@ -0,0 +1,2149 @@
 +/*
 + * Copyright (c) 2024 Gnattu OC <gnattuoc@me.com>
 + *
@@ -126,8 +126,7 @@ Index: FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
 +inline static float32x4_t mix_float32x4(float32x4_t x, float32x4_t y, float32x4_t a)
 +{
 +    float32x4_t n = vsubq_f32(y, x);
-+    n = vmulq_f32(n, a);
-+    n = vaddq_f32(n, x);
++    n = vfmaq_f32(x, n, a);
 +    return n;
 +}
 +
@@ -3831,7 +3830,7 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 +    out->color_trc = s->trc;
 +    out->colorspace = s->spc;
 +    out->color_primaries = s->pri;
-+    out->color_range = s->range;
++    if (s->range != -1) out->color_range = s->range;
 +
 +    if (in->color_trc == AVCOL_TRC_UNSPECIFIED)
 +        in->color_trc = AVCOL_TRC_SMPTE2084;
@@ -4339,7 +4338,7 @@ Index: FFmpeg/libavfilter/x86/vf_tonemapx_intrin_avx.c
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/x86/vf_tonemapx_intrin_avx.c
-@@ -0,0 +1,2277 @@
+@@ -0,0 +1,2276 @@
 +/*
 + * Copyright (c) 2024 Gnattu OC <gnattuoc@me.com>
 + *
@@ -4383,8 +4382,7 @@ Index: FFmpeg/libavfilter/x86/vf_tonemapx_intrin_avx.c
 +X86_64_V3 inline static __m128 mix_float32x4(__m128 x, __m128 y, __m128 a)
 +{
 +    __m128 n = _mm_sub_ps(y, x);
-+    n = _mm_mul_ps(n, a);
-+    n = _mm_add_ps(n, x);
++    n = _mm_fmadd_ps(a, n, x);
 +    return n;
 +}
 +
--- a/debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
+++ b/debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
@@ -95,7 +95,7 @@ Index: FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
-@@ -0,0 +1,1229 @@
+@@ -0,0 +1,2150 @@
 +/*
 + * Copyright (c) 2024 Gnattu OC <gnattuoc@me.com>
 + *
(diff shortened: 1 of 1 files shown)


debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
```diff
@@ -1263,1442 +1700,2454 @@ Index: FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
...
++    if (s->apply_dovi) {
++        AVFrameSideData *dovi_sd = av_frame_get_side_data(in, AV_FRAME_DATA_DOVI_METADATA);
++        if (dovi_sd) {
++            const AVDOVIMetadata *metadata = (AVDOVIMetadata *) dovi_sd->data;
++            const AVDOVIRpuDataHeader *rpu = av_dovi_get_header(metadata);
++            // only map dovi rpus that don't require an EL and has rpu profile == 0
++            // for performance reason we only want to do reshaping when absolutely needed
++            // such videos usually have vdr_rpu_profile == 0, for example profile 5 videos
++            // this could be wrong as there is no public documentation on this field
++            if (rpu->disable_residual_flag && rpu->vdr_rpu_profile == 0) {
```
nyanmisaka, line 3886: `rpu->vdr_rpu_profile == 0` This also seems to work for dav1 profile10.

debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
```diff
@@ -1263,1442 +1700,2454 @@ Index: FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
...
++                    goto fail;
++
++                ff_map_dovi_metadata(s->dovi, metadata);
++            }
++        }
++
++        if (s->dovi) {
++            if (desc->comp[2].plane == 1) {
++                av_log(s, AV_LOG_ERROR, "Input pixel format has to be yuv420p10 for Dolby Vision reshaping\n");
++                abort();
```
nyanmisaka: Use `av_assert0(0)` instead?

debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
```diff
@@ -1263,1442 +1700,2454 @@ Index: FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
...
++    td.odesc = odesc;
++    td.peak  = peak;
++    ff_filter_execute(ctx, s->filter_slice, &td, NULL,
++                      FFMIN(outlink->h >> FFMAX(desc->log2_chroma_h, odesc->log2_chroma_h), ff_filter_get_nb_threads(ctx)));
++
++    av_frame_free(&in);
++
++    av_frame_remove_side_data(out, AV_FRAME_DATA_MASTERING_DISPLAY_METADATA);
++    av_frame_remove_side_data(out, AV_FRAME_DATA_CONTENT_LIGHT_LEVEL);
++
```
nyanmisaka, line 3926:
```c
av_frame_remove_side_data(out, AV_FRAME_DATA_DOVI_RPU_BUFFER);
av_frame_remove_side_data(out, AV_FRAME_DATA_DOVI_METADATA);
```

debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
```diff
@@ -2714,331 +4163,1872 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 +    { "param",        "tonemap parameter", OFFSET(param), AV_OPT_TYPE_DOUBLE, {.dbl = NAN}, DBL_MIN, DBL_MAX, FLAGS },
 +    { "desat",        "desaturation strength", OFFSET(desat), AV_OPT_TYPE_DOUBLE, {.dbl = 0}, 0, DBL_MAX, FLAGS },
 +    { "peak",         "signal peak override", OFFSET(peak), AV_OPT_TYPE_DOUBLE, {.dbl = 0}, 0, DBL_MAX, FLAGS },
++    { "apply_dovi",  "Apply Dolby Vision metadata if possible", OFFSET(apply_dovi), AV_OPT_TYPE_BOOL, { .i64 = 1 }, 0, 1, FLAGS },
```
nyanmisaka, line 4167: `range` option default to `-1`?

debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
```diff
@@ -2714,331 +4163,1872 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
...
++#    if (__GNUC__ >= 9) || (__clang_major__ >= 11)
++#        define CC_SUPPORTS_TONEMAPX_INTRINSICS
++#    endif // (__GNUC__ >= 10) || (__clang_major__ >= 11)
++#endif // defined(__GNUC__) || defined(__clang__)
++
++#if defined(__clang__)
++#define UNROLL_LOOP #pragma clang loop unroll(full)
++#else
++#define UNROLL_LOOP GCC unroll n
++#endif
```
nyanmisaka: `UNROLL_LOOP` unused?

debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
gnattu: We don't have official support for profile10 on server side yet though.

Shadowghost (approved):

nyanmisaka:
With this change, it might be time to switch to using clang by default for win64, even though the static binary will be twice as large.

gnattu:
> With this change, it might be time to switch to using clang by default for win64, even though the static binary will be twice as large.

cc @joshuaboniface 

We might want to package the Clang-built FFmpeg on Windows starting with Jellyfin 10.10 and Jellyfin-FFmpeg 7.0. This might happen earlier with the unstable builds, but I'm not sure as we don't publish unstable ffmpeg binaries like server.

debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
```diff
@@ -123,6 +123,213 @@ Index: FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
 +#endif // ENABLE_TONEMAPX_NEON_INTRINSICS
 +
 +#ifdef ENABLE_TONEMAPX_NEON_INTRINSICS
++inline static float32x4_t mix_float32x4(float32x4_t x, float32x4_t y, float32x4_t a)
++{
++    float32x4_t n = vsubq_f32(y, x);
++    n = vmulq_f32(n, a);
++    n = vaddq_f32(n, x);
```
Bond-009: Why not use FMA here?

debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
nyanmisaka:
```
[tonemapx @ 0000018213ccaa40] Unsupported output color range -1 ((null))
```
Oops. This needs to be handled internally.

gnattu: Handled by [6d0b354](https://github.com/jellyfin/jellyfin-ffmpeg/pull/447/commits/6d0b354ec5ad32772eea70bf9a4f6ca15610a91f)

nyanmisaka (approved):

PR #451, opened 2024-09-06
Fix ocl tonemap cannot output full range

nyanmisaka:
**Changes**
- Fix ocl tonemap cannot output full range

**Issues**
```diff
 float rgb2y(float r, float g, float b) {
     float y = r*yuv_matrix[0] + g*yuv_matrix[1] + b*yuv_matrix[2];
+#ifndef FULL_RANGE_OUT
     y = (219.0f * y + 16.0f) / 255.0f;
+#endif
     return y;
 }
```


gnattu (approved):
I've noticed this for Metal before and fixed that earlier, did not check OCL.

nyanmisaka:
> I've noticed this for Metal before and fixed that earlier, did not check OCL.

I also fixed it in CUDA in the past and forgot to add it back into OCL.
The problem was not exposed until the recent changes.

Shadowghost (approved):

PR #454, opened 2024-09-08
lavc/videotoolbox: allow software fallback for all codecs

gnattu:
Apple disabled hardware decoding for some h264 files with certain condition and now ffmpeg will error out for such inputs because the software fallback is disabled. Allow software fallback for all codecs instead of only for HEVC to workaround this as the error handling outside ffmpeg would be harder. Allowing software fallback has no measurable performance impact when the hardware decoder is not overloaded.

Refer: macOS 14.6 release note: https://developer.apple.com/documentation/macos-release-notes/macos-14_6-release-notes#Video-Toolbox

The impacted videos are rare but still exist. 

**Changes**

**Issues**

### DIFF
--- a/debian/patches/0071-allow-vt-sw-decoder-for-every-codec.patch
+++ b/debian/patches/0071-allow-vt-sw-decoder-for-every-codec.patch
@@ -1,0 +1,15 @@
+Index: FFmpeg/libavcodec/videotoolbox.c
+===================================================================
+--- FFmpeg.orig/libavcodec/videotoolbox.c
++++ FFmpeg/libavcodec/videotoolbox.c
+@@ -812,9 +812,7 @@ static CFDictionaryRef videotoolbox_deco
+                                                                    &kCFTypeDictionaryValueCallBacks);
+ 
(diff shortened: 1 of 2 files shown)


nyanmisaka (commented):
```shell
echo 0071-allow-vt-sw-decoder-for-every-codec.patch >> debian/patches/series
```

nyanmisaka:
In this case, which one is faster, the software fallback provided by Apple or the built-in decoder in ffmpeg?

gnattu:
> ```shell
> echo 0071-allow-vt-sw-decoder-for-every-codec.patch >> debian/patches/series
> ```

Forget add that to the commit, re-pushed

gnattu:
> In this case, which one is faster, the software fallback provided by Apple or the built-in decoder in ffmpeg?

FFmpeg native is (much) faster in benchmark but there should not have impact on real world scenario because the VT sw decoder seems to be throttled as only two P cores are in use and each only have around 50% usage, while FFmpeg native used up all cores.

nyanmisaka (approved):

PR #457, opened 2024-09-12
avfilter: add vf_bwdif_videotoolbox

gnattu:
**Changes**

This ports back the bwdif metal implementation from handbrake: https://github.com/HandBrake/HandBrake/blob/master/libhb/platform/macosx/shaders/bwdif_vt.metal, which is a port of ffmpeg's `vf_bwdif_cuda`.

The filter main file is very similar to the `yadif` variant with only slight modification to pass in bwdif specific parameters. 

**Issues**


Shadowghost (approved):

nyanmisaka (approved):
LGTM. BTW handbrake also has some ffmpeg patches that might be useful to us, though I haven't found a use case to test.

PR #459, opened 2024-09-18
hevc: add 12bit videotoolbox decoding support

gnattu:
**Changes**

12bit HEVC works with minimal changes if we process all filters using downsampled `yuv420p10`. However, it's important to note that Intel Macs never had hardware support for 12-bit processing, so they will always fall back on software decoding. Even on M1 series, hardware acceleration for 12-bit is incomplete and likely uses hybrid decoding. This is evident by the higher CPU usage compared to 10-bit 4:4:4 inputs, though the M1 still achieves better framerates with lower CPU usage than FFmpeg's native software decoder when decoding 4K HDR 12-bit 4:4:4 inputs.

**Issues**

### DIFF
--- a/debian/patches/0073-add-12bit-decoding-on-videotoolbox.patch
+++ b/debian/patches/0073-add-12bit-decoding-on-videotoolbox.patch
@@ -1,0 +1,24 @@
+Index: FFmpeg/libavcodec/hevcdec.c
+===================================================================
+--- FFmpeg.orig/libavcodec/hevcdec.c
++++ FFmpeg/libavcodec/hevcdec.c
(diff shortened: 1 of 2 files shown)


nyanmisaka (approved):
Interestingly, the three vendors use different pixel memory layouts for yuv444p12.

AV_PIX_FMT_XV36 <- Intel
[AV_PIX_FMT_P416](https://github.com/FFmpeg/FFmpeg/commit/f0952f87bcfd905f21ce6cefa8e14a127fb00c8a) <- Apple
AV_PIX_FMT_YUV444P16 <- NVIDIA

PR #462, opened 2024-09-20
ffprobe: add show_vt_info option

gnattu:
**Changes**

This adds a new option `-show_vt_info` to ffprobe when VideoToolbox is configured at build time, which output decoder and encoder info in a minimized json. For example:

```
./ffprobe -hide_banner -show_vt_info
{"Decoders":["AV_CODEC_ID_H264","AV_CODEC_ID_HEVC","AV_CODEC_ID_VP9"],"Encoders":[{"Codec":"AV_CODEC_ID_H264","MaxWidth":4096,"Profiles":["FF_PROFILE_H264_BASELINE","FF_PROFILE_H264_CONSTRAINED_BASELINE","FF_PROFILE_H264_MAIN","FF_PROFILE_H264_HIGH","FF_PROFILE_H264_HIGH_422","FF_PROFILE_H264_HIGH_444_PREDICTIVE"],"SupportYuv444Encode":true,"MaxHeight":4096,"Support10bitEncode":true,"SupportHDREncode":true},{"Codec":"AV_CODEC_ID_HEVC","MaxWidth":8192,"Profiles":["FF_PROFILE_HEVC_MAIN","FF_PROFILE_HEVC_MAIN_10","FF_PROFILE_HEVC_MAIN_STILL_PICTURE","FF_PROFILE_HEVC_REXT"],"SupportYuv444Encode":true,"MaxHeight":8192,"Support10bitEncode":true,"SupportHDREncode":true},{"SupportHDREncode":false,"MaxWidth":8192,"Support10bitEncode":false,"MaxHeight":8192,"SupportYuv444Encode":false,"Codec":"AV_CODEC_ID_MJPEG"}]}
```

For decoders, only the codec name where the system declares hardware acceleration is available is printed, with no extra information. This is because:

- There isn't a straightforward API to fetch detailed information for decoders. You would need to emulate a real video input, including extra data like NAL units, to check if the decoding session can be created with software fallback disabled.

- VideoToolbox gracefully handles software fallback internally, making such detailed checks redundant for our use cases.

For encoders, extra data like the max dimension supported is also printed for us to pick encoders in the future.

I prefer adding this option to ffprobe rather than creating another standalone fftool. Creating an additional portable binary just for hardware info (which would be around 50-80MB) seems excessive to me.

We also need to standardize the JSON output format for hardware info across different platforms. The current draft for other hardware exposes too much unnecessary information and isn't very easy to work with. 

**Issues**

### DIFF
--- a/debian/patches/0075-ffprobe-add-vt-info.patch
+++ b/debian/patches/0075-ffprobe-add-vt-info.patch
@@ -357,3 +357,27 @@ Index: FFmpeg/fftools/opt_vtinfo.m
 +    }
 +    return 0;
 +}
+Index: FFmpeg/fftools/ffprobe.c
+===================================================================
+--- FFmpeg.orig/fftools/ffprobe.c
++++ FFmpeg/fftools/ffprobe.c
+@@ -69,6 +69,9 @@
+ #include "opt_common.h"
+ 
+ #include "libavutil/thread.h"
++#ifdef CONFIG_VIDEOTOOLBOX
++    #include "opt_vtinfo.h"
++#endif
+ 
+ #if !HAVE_THREADS
+ #  ifdef pthread_mutex_lock
+@@ -4587,6 +4590,9 @@ static const OptionDef real_options[] =
+     { "print_filename",        OPT_TYPE_FUNC, OPT_FUNC_ARG, {.func_arg = opt_print_filename}, "override the printed input filename", "print_file"},
+     { "find_stream_info",      OPT_TYPE_BOOL, OPT_INPUT | OPT_EXPERT, { &find_stream_info },
+         "read and decode the streams to fill missing information with heuristics" },
++#ifdef CONFIG_VIDEOTOOLBOX
++    { "show_vt_info",          OPT_TYPE_FUNC, OPT_EXIT, { .func_arg = show_vt_info }, "show VideoToolbox info" },
++#endif
+     { NULL, },
+ };
+ 
--- a/debian/patches/0075-ffprobe-add-vt-info.patch
+++ b/debian/patches/0075-ffprobe-add-vt-info.patch
@@ -1,0 +1,359 @@
+Index: FFmpeg/fftools/Makefile
+===================================================================
+--- FFmpeg.orig/fftools/Makefile
++++ FFmpeg/fftools/Makefile
+@@ -38,6 +38,10 @@ $(1)$(PROGSSUF)_g$(EXESUF): FF_EXTRALIBS
+ -include $$(OBJS-$(1):.o=.d)
+ endef
+ 
++ifdef CONFIG_VIDEOTOOLBOX
++  OBJS-ffprobe += fftools/opt_vtinfo.o
++endif
++
+ $(foreach P,$(AVPROGS-yes),$(eval $(call DOFFTOOL,$(P))))
+ 
+ ifdef HAVE_GNU_WINDRES
+Index: FFmpeg/fftools/opt_vtinfo.h
+===================================================================
+--- /dev/null
++++ FFmpeg/fftools/opt_vtinfo.h
+@@ -0,0 +1,6 @@
++/**
++ * Print the VideoToolbox Hardware decoder/encoder info
++ * For decoders, only the decoder name the system claims to support will be printed
++ * This option processing function does not utilize the arguments.
++ */
++int show_vt_info(void *optctx, const char *opt, const char *arg);
+Index: FFmpeg/fftools/opt_vtinfo.m
+===================================================================
+--- /dev/null
++++ FFmpeg/fftools/opt_vtinfo.m
+@@ -0,0 +1,328 @@
++/*
++ * Copyright (c) 2024 Gnattu OC
++ *
++ * This file is part of FFmpeg.
++ *
++ * FFmpeg is free software; you can redistribute it and/or
++ * modify it under the terms of the GNU Lesser General Public
++ * License as published by the Free Software Foundation; either
++ * version 2.1 of the License, or (at your option) any later version.
++ *
++ * FFmpeg is distributed in the hope that it will be useful,
++ * but WITHOUT ANY WARRANTY; without even the implied warranty of
++ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
++ * Lesser General Public License for more details.
++ *
++ * You should have received a copy of the GNU Lesser General Public
++ * License along with FFmpeg; if not, write to the Free Software
++ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
++ */
++
++#import <Foundation/Foundation.h>
++#import <VideoToolbox/VideoToolbox.h>
++#import <CoreMedia/CoreMedia.h>
++#include "opt_vtinfo.h"
++
++CMVideoCodecType decoderCodecs[] = {
++    kCMVideoCodecType_MPEG2Video,
++    kCMVideoCodecType_MPEG4Video,
++    kCMVideoCodecType_H264,
++    kCMVideoCodecType_HEVC,
++    kCMVideoCodecType_VP9,
++    kCMVideoCodecType_AV1,
++};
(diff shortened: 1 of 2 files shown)


Shadowghost (dismissed):

nyanmisaka:
Our lives would be much easier if Linux and Windows could also provide software fallbacks like Apple does, but unfortunately this is not going to happen.

> There isn't a straightforward API to fetch detailed information for decoders. You would need to emulate a real video input, including extra data like NAL units, to check if the decoding session can be created with software fallback disabled.

We choose to trust the capabilities reported by the driver. But if it fails (which is likely to happen especially with Intel GPUs that lack firmware), I'd rather handle software fallback in Jellyfin.

> I prefer adding this option to ffprobe rather than creating another standalone fftool. Creating an additional portable binary just for hardware info (which would be around 50-80MB) seems excessive to me.

I originally wanted to put the code in `ffprobe` as well, but I found that I didn't want to disturb ffprobe's existing logic. And just like the overhaul of `ffmpeg.c` in ffmpeg 7.0, almost any large patches to `ffmpeg.c` no longer apply.

```
  --disable-all            disable building components, libraries and programs
```

You can simply using `--disable-all --enable-fftool...` for this standalone fftool, or best, fix and build with `--enable-shared`.

> We also need to standardize the JSON output format for hardware info across different platforms. The current draft for other hardware exposes too much unnecessary information and isn't very easy to work with.

I agree to unify common attributes of JSON output, but some HWA/vendor specific attributes are also needed, such as driver version.

nyanmisaka:
Another problem is what json writer to use. There is no NSJSONSerialization on platforms other than Apple, so we have to use the json utils provided by ffmpeg. But the data format must be the same on all platforms.

gnattu:
> There is no NSJSONSerialization on platforms other than Apple

VideoToolbox info is specifically for Apple platforms so that is fine to use that as it is easier to work with than the ffmpeg one. As long as the serialized json has the same schema I think we should be fine.

> I originally wanted to put the code in ffprobe as well, but I found that I didn't want to disturb ffprobe's existing logic.

The change to original file is quite small though? You just push one more option into the ffprobe.c, adding more includes, and all other logics are handled in our own files which makes the conflicts easy to be resolved. [44cd30b](https://github.com/jellyfin/jellyfin-ffmpeg/pull/462/commits/44cd30baef0f4eb336ec54d3840b1e77c4a11550)

> You can simply using --disable-all --enable-fftool... for this standalone fftool, or best, fix and build with --enable-shared.

If we can disable all other things we can just not depends on ffmpeg at all and make a lightweight standalone tool ourselves though? Fix enable-shared is not trivial due to ffmpeg's weird build system and clang's special linkers. The portable builds are expected to be fully static though, and windows was the only outlier in the history. Is there anything that ffmpeg provides other than type enums that has to be used here? 

> But if it fails (which is likely to happen especially with Intel GPUs that lack firmware), I'd rather handle software fallback in Jellyfin.

This requires a transcoder rewrite though. We catch the exit status and retry with pure software. Some problematic hardware does not even fail the pipeline but just output wrong pictures, and I think that will require admin actions.

nyanmisaka:
> VideoToolbox info is specifically for Apple platforms so that is fine to use that as it is easier to work with than the ffmpeg one. As long as the serialized json has the same schema I think we should be fine.

I guess you could mix NSJSON with ffmpeg's json writer. But if you don't use ffmpeg's writer at all, the output might be hard to handle, and you'd need to manually manage [the sections](https://github.com/nyanmisaka/jellyfin-ffmpeg/blob/540d6aab375e1d22d0b6a22f8feffdf1135eace8/fftools/ffhwinfo_utils.c#L30) to be consistent with other platforms. And the output is limited to json. ffprobe's [default format](https://www.ffmpeg.org/ffprobe.html#default) is not supported.

> The change to original file is quite small though? You just push one more option into the ffprobe.c, adding more includes, and all other logics are handled in our own files which makes the conflicts easy to be resolved. [44cd30b](https://github.com/jellyfin/jellyfin-ffmpeg/pull/462/commits/44cd30baef0f4eb336ec54d3840b1e77c4a11550)

It should be feasible to just add a few new options with `OPT_EXIT` flag to ffprobe.

> If we can disable all other things we can just not depends on ffmpeg at all and make a lightweight standalone tool ourselves though? Fix enable-shared is not trivial due to ffmpeg's weird build system and clang's special linkers. The portable builds are expected to be fully static though, and windows was the only outlier in the history. Is there anything that ffmpeg provides other than type enums that has to be used here?

Then it will be a problem to distribute this standalone tool downstream, such as the Archlinux repo which builds from source. And there is no need to create a new package for this. In addition, some headers and compat(dyn linked cuda/nvml) libs require ffmpeg/configure.

> This requires a transcoder rewrite though. We catch the exit status and retry with pure software. Some problematic hardware does not even fail the pipeline but just output wrong pictures, and I think that will require admin actions.

In the worst case the `h26x_qsv` decoder cannot find the sequence header in the bitstream, which causes it to keep trying the next packet and not exit gracefully.

gnattu:
> I guess you could mix NSJSON with ffmpeg's json writer. 

It is not that easy as the data type inside is completely different

> But if you don't use ffmpeg's writer at all, the output might be hard to handle, and you'd need to manually manage [the sections](https://github.com/nyanmisaka/jellyfin-ffmpeg/blob/540d6aab375e1d22d0b6a22f8feffdf1135eace8/fftools/ffhwinfo_utils.c#L30) to be consistent with other platforms.

Current sections are not suitable for VT specific cases anyway. The consistency is easier to handle than you think after we decided on how the output json schema would look like.

> And the output is limited to json. ffprobe's [default format](https://www.ffmpeg.org/ffprobe.html#default) is not supported.

I think json only output is Okay for our use case, this is not a general cli tool for human read, and json is human friendly enough after prettify.

gnattu:
It turns out that VT does have some private APIs that can get more info about the decoder and those APIs are used by Safari.

Those are not exposed directly in `VideoToolbox.h` but it is callable by define an `extern` function signature.

For example: `VTCopyHEVCDecoderCapabilitiesDictionary` would return something like:

```json
{
  "VTIsHDRAllowedOnDevice": true,
  "VTSupportedProfiles": [
    1,
    2,
    3,
    4
  ],
  "VTDoViIsHardwareAccelerated": true,
  "VTDoViSupportedLevels": [
    "01",
    "02",
    "03",
    "04",
    "05",
    "06",
    "07",
    "08",
    "09"
  ],
  "VTPerProfileSupport": {
    "1": {
      "VTMaxPlaybackLevel": 186,
      "VTIsHardwareAccelerated": true,
      "VTMaxDecodeLevel": 186
    },
    "2": {
      "VTMaxPlaybackLevel": 186,
      "VTIsHardwareAccelerated": true,
      "VTMaxDecodeLevel": 186
    },
    "3": {
      "VTMaxPlaybackLevel": 186,
      "VTIsHardwareAccelerated": true,
      "VTMaxDecodeLevel": 186
    },
    "4": {
      "VTMaxPlaybackLevel": 186,
      "VTIsHardwareAccelerated": true,
      "VTMaxDecodeLevel": 186
    }
  },
  "VTDoViSupportedProfiles": [
    "05",
    "08"
  ]
}
```

The profile here is the number representation of HEVC general level idc, where `1 == Main`, `2 == Main10`, `3 == MainStill` and `4 == RExt`. 

There is also a similar method for AV1: `VTCopyAV1DecoderCapabilitiesDictionary`, but that one returns null on M1 series. 

I'm not very confident to use private APIs like this, and probably we don't need such info as:
- VT fallbacks to software gracefully
- The DoVi related capabilities are not fully exposed

nyanmisaka:
> I'm not very confident to use private APIs like this, and probably we don't need such info as:
> - VT fallbacks to software gracefully
> - The DoVi related capabilities are not fully exposed

Due to the existence of software fallback, Level info seems unnecessary. In addition, DoVi metadata is parsed in ffmpeg, and we only need to decode HEVC bitstream/base layer from VT.

gnattu:
Replaced by [#19](https://github.com/nyanmisaka/jellyfin-ffmpeg/pull/19)

PR #468, opened 2024-09-28
Enable VC1 decode on Intel Arc dGPU

nyanmisaka:
**Changes**
- Enable VC1 decode on Intel Arc dGPU (MTL+ is not supported)
- Update dependencies

**Issues**
- https://github.com/jellyfin/jellyfin/issues/9588#issuecomment-1496031539

### DIFF
--- a/builder/scripts.d/10-mingw.sh
+++ b/builder/scripts.d/10-mingw.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 SCRIPT_REPO="https://git.code.sf.net/p/mingw-w64/mingw-w64.git"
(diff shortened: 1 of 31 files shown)


gnattu:
So this is only available on Linux?

nyanmisaka:
> So this is only available on Linux?

On Windows it has been supported since day one. On Linux there are very few VC1 clients. And DG2's next gen MTL completely removed VC1 support.

Shadowghost (approved):

nyanmisaka:
For the record, the VC1 software decoder in ffmpeg only supports using a single thread.

PR #470, opened 2024-10-01
Fix hardcoded color range in tonemapx LUTs

nyanmisaka:
**Changes**
- Fix hardcoded color range in tonemapx LUTs
- Fix typo and clean up code to be consistent with GPGPU filters

**Issues**
- Obviously different colors between sw tonemapx and GPGPU filters (in maxRGB)

### DIFF
--- a/debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
+++ b/debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
@@ -95,7 +95,7 @@ Index: FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
-@@ -0,0 +1,2149 @@
+@@ -0,0 +1,2150 @@
 +/*
 + * Copyright (c) 2024 Gnattu OC <gnattuoc@me.com>
 + *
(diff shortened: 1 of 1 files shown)


gnattu:
I have a local version to make more changes to keep GPU/CPU implementations consistent to replace this because our GPU implementation side also has some inconsistencies abut range and peak handling that is different from the software one. But I noticed some design differences in yours so maybe we need to discuss further.

debian/patches/0060-add-simd-optimized-tonemapx-filter.patch
```diff
@@ -2937,12 +2943,14 @@ Index: FFmpeg/libavfilter/vf_tonemapx.c
 +        return AVERROR(ENOMEM);
 +    if (!s->delin_lut && !(s->delin_lut = av_calloc(32768, sizeof(uint16_t))))
 +        return AVERROR(ENOMEM);
++    if (s->lut_delin_max <= 0)
++        return AVERROR(EINVAL);
 +
 +    for (i = 0; i < 32768; i++) {
-+        double v1 = (i - 2048.0f) / 28672.0f;
++        double v1 = (i - s->lut_off) / 32767.0f;
```
gnattu, line 2950: Why it is not divided by limited range `28672.0f` when the input is in limited range now? In my implementation this is changed with the `lut_off`.

nyanmisaka: Because doing so gives incorrect colors based on testing, too bright and loses too much detail.

off
fine

gnattu: This is due to still incorrect range handling though, as the output rgb color after yuv2rgb is always in full range this limited range lut should not be used on those colors. I'm going to push my version soon.

PR #484, opened 2024-10-21
avformat: validate dovi config in muxers

gnattu:
FFmpeg tends to copy any available side packets from the input streams into the muxed output without considering their validity. This behavior is problematic for Dolby Vision configuration records, as invalid records might crash the player. I haven't found a better place to perform such validations with the current Dolby Vision handling process, so I had to place the validation code into each muxers.

This PR checks whether the Dolby Vision record is valid, meaning it is spec-compliant, so the player will likely handle it without issues in `movenc`, `mpegtsenc`, and `matroskaenc`. It first verifies if the codec matches the profile, then checks if the base layer (BL) is encoded as required by the compatibility ID. If the compatibility ID is 0, it also checks if the codec tag is explicitly set to the dovi fourcc, as these videos will not have backward compatibility.

The check in `mpegtsenc` is slightly stripped down as we are only passing HEVC with BL compat in that muxer.

**Changes**

**Issues**

Fixes #483

PR #485, opened 2024-10-21
avcodec/audiotoolboxenc: fix 7.1 channel mapping and sample rate

gnattu:
**Changes**

Both `AV_CHANNEL_LAYOUT_7POINT1_WIDE_BACK` and `AV_CHANNEL_LAYOUT_7POINT1`
should be mapped to `kAudioChannelLayoutTag_AAC_7_1_B`, otherwise neither
the back surround channel nor the side surround channel has sound.

aac_at has stricter sample rate than other encoders which should be
spcified to enable auto resampling.

alac with sample rate higher than 38400 will not play on Apple devices,
although higher sample rates are accepted by the encoder

All encoders provided by AudioToolbox can handle 32bit integer samples,
enable that for all encoders.

This patch is probably upstream-able, will submit later.

**Issues**

This includes #484 to avoid conflicts.  Marks as draft until that one is merged.

### DIFF
--- a/debian/patches/0078-fix-atenc-layout-samplerate.patch
+++ b/debian/patches/0078-fix-atenc-layout-samplerate.patch
@@ -1,0 +1,67 @@
+Index: FFmpeg/libavcodec/audiotoolboxenc.c
+===================================================================
+--- FFmpeg.orig/libavcodec/audiotoolboxenc.c
++++ FFmpeg/libavcodec/audiotoolboxenc.c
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

gnattu:
Upstreaming requires a bit more work to use a modernized api `get_supported_config` instead.

nyanmisaka (approved):

Bond-009 (approved):

PR #488, opened 2024-10-23
lavc/videotoolbox: remove opengl compatability key

gnattu:
We are not using OpenGL and this compatability key could introduce performance penalty on some Macs. Performance difference is neglible on Apple Silicon but on old Intel Macs like the 2018 MacBook Air, decoding performance could be reduced by 15% as the IOSurface might make extra frame copies to reshape the pixel buffer.

**Changes**

**Issues**

### DIFF
--- a/debian/patches/0079-videotoolbox-remove-opengl-compatability.patch
+++ b/debian/patches/0079-videotoolbox-remove-opengl-compatability.patch
@@ -1,0 +1,16 @@
+Index: FFmpeg/libavcodec/videotoolbox.c
+===================================================================
+--- FFmpeg.orig/libavcodec/videotoolbox.c
++++ FFmpeg/libavcodec/videotoolbox.c
+@@ -786,11 +786,6 @@ static CFDictionaryRef videotoolbox_buff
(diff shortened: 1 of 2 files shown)


nyanmisaka (approved):

nyanmisaka:
`kCVPixelBufferMetalCompatibilityKey` Can this key bring any benefits in Metal?

gnattu:
> `kCVPixelBufferMetalCompatibilityKey` Can this key bring any benefits in Metal?

That key is not required as long as the pixel buffer is `IOSurface` backed and that key is always set:

https://github.com/jellyfin/jellyfin-ffmpeg/blob/982e2b8024537a157ffb0c2021e7ddafb37c296f/libavcodec/videotoolbox.c#L786

For all `IOSurface` backed `CVPixelBuffer` the `MTLTexture` created by `CVMetalTextureCacheCreateTextureFromImage` will use the `CVPixelBuffer` itself for texture storage which means no GPU frame uploading or extra texture allocation is required, at least not on Apple GPUs.

Bond-009 (approved):

Shadowghost (approved):

Scutua (approved):

PR #500, opened 2024-11-15
avcodec/videotoolbox: backport AV1 hardware acceleration

gnattu:
**Changes**

This backports AV1 hardware acceleration from upstream and has been tested on the M4 Max. It is important to note that unlike other hardware acceleration provided by VideoToolbox, Apple does not offer any software fallback for AV1. Consequently, the decode will silently fail for all frames when the system lacks AV1 support or when the video utilizes parameters not supported by the decoder (e.g., 4:2:2 and 4:4:4 color). To address this issue, we must implement a server-side workaround to provide graceful software fallback. This is feasible as Apple has a finite set of chips that do not support AV1, while all current chips support AV1 and possess the same decoding capabilities. 

**Issues**

### DIFF
--- a/debian/patches/0081-backport-av1-videotoolbox.patch
+++ b/debian/patches/0081-backport-av1-videotoolbox.patch
@@ -1,0 +1,352 @@
+Index: FFmpeg/libavcodec/Makefile
+===================================================================
+--- FFmpeg.orig/libavcodec/Makefile
++++ FFmpeg/libavcodec/Makefile
+@@ -1010,6 +1010,7 @@ OBJS-$(CONFIG_AV1_D3D12VA_HWACCEL)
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):
I'm kind of astonished that 422 and 444 are not supported.

gnattu:
> I'm kind of astonished that 422 and 444 are not supported.

It is normal as those are not supported by current gen hardware from all vendors.

nyanmisaka:
Dav1d is already highly optimized on ARM via NEON (+some SVE(2)), not sure why Apple only provides AVIF support via it.

nyanmisaka (approved):

Bond-009 (approved):

PR #503, opened 2024-11-18
builder: use github as libisl mirror for ct-ng

gnattu:
The old sourceforge mirror is not reliable and ct-ng is recommending using github as the mirror for libisl

**Changes**

**Issues**


gnattu:
Closed as the mirror is wrong

PR #504, opened 2024-11-18
builder: use github as libisl mirror for ct-ng

gnattu:
The previously used official SourceForge mirror is not reliable as the upstream hosting appears to be IP-blocking some GitHub runners’ IP addresses. Replace it with a self-hosted mirror on GitHub. This mirror is a repository created specifically for this purpose and can be updated if needed.

**Changes**

**Issues**

### DIFF
--- a/builder/images/base-linux64/ct-ng-config
+++ b/builder/images/base-linux64/ct-ng-config
@@ -825,7 +825,7 @@ CT_ISL_V_0_26=y
 # CT_ISL_V_0_16 is not set
 # CT_ISL_V_0_15 is not set
 CT_ISL_VERSION="0.26"
-CT_ISL_MIRRORS="https://libisl.sourceforge.io"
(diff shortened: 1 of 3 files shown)


Shadowghost (approved):

nyanmisaka:
I tried looking in the logs but couldn't find any errors related to `libisl`. Can you attach the relevant logs?

gnattu:
> I tried looking in the logs but couldn't find any errors related to `libisl`. Can you attach the relevant logs?

It is from #501 where some CI nodes get 403 on downloading libisl. The problem here is that our CI took a very long time to finish and random error due to such issue is very very annoying. This is not the first time that our CI failed with package downloading issues and I'm not a big fan of it. Handbrake did something similar that archives all packages used in their own repo: https://github.com/HandBrake/HandBrake-contribs

nyanmisaka (approved):

PR #509, opened 2024-11-26
avformat/assenc: do not copy null terminator

### LINKED ISSUE #506: ASS Sub BUG
> As I wrote above, the problem is solved by manually extracting the ASS subtitles and adding them without changing them with the MKVTool and recreating the file.
> 
> For a large collection, this would mean, in my case, several 10,000 files that would have to be edited manually although it was once fixed and worked. With the update to version 10 this was broken again. 
> 
> To be honest, I have no desire to edit 10,000+ files manually as there is obviously another way. 
> 
> If this is the wrong section, I would ask you to move this so that the error can be addressed. Thank you
> 
>  

 _Originally posted by @Snipy2k4 in [#267](https://github.com/jellyfin/jellyfin-webos/issues/267#issuecomment-2481237254)_

gnattu:
The `par->extradata` buffer filled from some matroska files may be null terminated, and use `ffio_write_lines` using the full buffer length will copy this null character into the output files. This results in a file in which there is a null terminator after the header, but preceeding the actual content of the subtitle file. Treat this buffer as a string and write line with the text length of this buffer to skip the null character.

Regression from 7bf1b9b

Fixes #506 

**Changes**

**Issues**

### DIFF
--- a/debian/patches/series
+++ b/debian/patches/series
@@ -79,3 +79,4 @@
 0079-videotoolbox-remove-opengl-compatability.patch
 0080-use-dynamic-pool-for-vpl-qsv-hwupload.patch
 0081-backport-av1-videotoolbox.patch
+0082-fix-ass-incorrect-null-copy.patch
--- a/debian/patches/0082-fix-ass-incorrect-null-copy.patch
+++ b/debian/patches/0082-fix-ass-incorrect-null-copy.patch
@@ -1,0 +1,14 @@
+Index: FFmpeg/libavformat/assenc.c
+===================================================================
+--- FFmpeg.orig/libavformat/assenc.c
++++ FFmpeg/libavformat/assenc.c
+@@ -67,7 +67,8 @@ static int write_header(AVFormatContext
+                 ass->trailer = trailer;
+         }
+ 
+-        ffio_write_lines(s->pb, par->extradata, header_size, NULL);
++        header_size = av_strnlen(par->extradata, header_size);
++        ffio_write_lines(s->pb, par->extradata, (int)header_size, NULL);
+ 
+         ass->ssa_mode = !strstr(par->extradata, "\n[V4+ Styles]");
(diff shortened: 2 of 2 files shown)


debian/patches/series
```diff
@@ -79,3 +79,4 @@
 0079-videotoolbox-remove-opengl-compatability.patch
 0080-use-dynamic-pool-for-vpl-qsv-hwupload.patch
 0081-backport-av1-videotoolbox.patch
+0082-fix-ass-incorrect-null-copy
```
nyanmisaka: Missing .patch suffix in two files.

```suggestion
0082-fix-ass-incorrect-null-copy.patch
```

nyanmisaka:
If this is the cause, it would be nice to have the upstream add a sample to their fate-tests.

gnattu:
The problem is I'm unable to reproduce with any of the files and handbrake combinations, the sample file I used for testing is binary edited to contain the null char. The sample file in the original ticket is no longer downloadable.

nyanmisaka:
> The problem is I'm unable to reproduce with any of the files and handbrake combinations, the sample file I used for testing is binary edited to contain the null char. The sample file in the original ticket is no longer downloadable.

Exactly. Using those short links is terrible for archiving.
I also tried searching locally but it seems I've deleted it because it's been so long.

Shadowghost (approved):

gnattu:
Merging as confirmed to fix the issue

ethaldeman:
@nyanmisaka @gnattu I don't know if this topic is of concern or interest anymore, but I created a sample mkv with the null character problem for https://github.com/peerless2012/libass-android/issues/55 . 
[broken_ass_subtitles.zip](https://github.com/user-attachments/files/22547904/broken_ass_subtitles.zip)
I can also reliability reproduce the bug using Handbrake 1.9.2 and likely any older version.

PR #527, opened 2025-01-19
Fix a race condition in VAAPI CSC to YUV420P

nyanmisaka:
**Changes**
- Fix a race condition in VAAPI CSC to YUV420P
  ```
  There is a race condition for AV_PIX_FMT_YUV420P when mapping from pix_fmt
  to fourcc, both VA_FOURCC_I420 and VA_FOURCC_YV12 could be found by pix_fmt.

  fixes "-vf scale_vaapi=format=yuv420p"
  ```
- Sync fixes from ffmpeg-rockchip
  ```
  * refine hw_device_ctx validation
  * enable some YUVJ formats.
  ```
- Update dependencies
- Bump version to 7.0.2-9


Shadowghost (approved):

gnattu:
libx265 introduced a regression in commit https://bitbucket.org/multicoreware/x265_git/commits/f5e4a648727ea86828df236eb44742fe3e3bf366#Lsource/common/x86/cpu-a.asmT181

It assumes that all Unix64 targets are Linux-like, using ELF binaries which have `..plt` (Procedure Linkage Table). This assumption is invalid for Mach-O binaries, which macOS uses, and causing compiler errors.

nyanmisaka:
> libx265 introduced a regression in commit https://bitbucket.org/multicoreware/x265_git/commits/f5e4a648727ea86828df236eb44742fe3e3bf366#Lsource/common/x86/cpu-a.asmT181
> 
> It assumes that all Unix64 targets are Linux-like, using ELF binaries which have `..plt` (Procedure Linkage Table). This assumption is invalid for Mach-O binaries, which macOS uses, and causing compiler errors.

Good catch. I think we should stay on 4.1 stable for now.

gnattu (approved):

PR #538, opened 2025-02-21
Disable x265 Neon_Dotprod+ ASM on Linux/Arm64 for now

nyanmisaka:
**Changes**
- Disable x265 Neon_Dotprod+ ASM on Linux/Arm64 for now

  This caused crashes on Cortex-A76/A55 CPUs with armv8.2-dotprod support.
  Ref: https://bitbucket.org/multicoreware/x265_git/issues/984/illegal-instruction-neon_dotprod-crashes

- Target armv8-a for Linux/Arm64 portable builds build

### DIFF
--- a/builder/scripts.d/50-x265.sh
+++ b/builder/scripts.d/50-x265.sh
@@ -28,6 +28,17 @@ ffbuild_dockerbuild() {
         )
     fi
 
+    # Wa for https://bitbucket.org/multicoreware/x265_git/issues/984/illegal-instruction-neon_dotprod-crashes
+    if [[ $TARGET == linuxarm64 ]]; then
+        common_config+=(
+            -DENABLE_NEON=ON
+            -DENABLE_NEON_DOTPROD=OFF
+            -DENABLE_NEON_I8MM=OFF
+            -DENABLE_SVE=OFF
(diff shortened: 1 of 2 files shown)


builder/scripts.d/50-x265.sh
```diff
@@ -20,6 +20,11 @@ ffbuild_dockerbuild() {
         -DENABLE_CLI=OFF
         -DCMAKE_ASM_NASM_FLAGS=-w-macro-params-legacy
         -DENABLE_ALPHA=ON
+        -DENABLE_NEON=ON
+        -DENABLE_NEON_DOTPROD=OFF
```
gnattu: These are common flags which will disable dotprod and i8mm on macOS builds too. We need to move those to the `[[ $TARGET != mac* ]]` sections below

gnattu (changes requested):

builder/scripts.d/50-x265.sh
nyanmisaka: oops i forgot macos builder also shared this

gnattu (approved):

Shadowghost (approved):

Issue #544, opened 2025-03-03
Accomodate SVT-AV1 API change


regularhunter:
SVT-AV1 3.0.0 changed the signature of `svt_av1_enc_init_handle`. ffmpeg has not released a version including the needed change.

https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/d1ed5c06e3edc5f2b5f3664c80121fa55b0baa95

```
author	Gyan Doshi <ffmpeg@gyani.pro>	
	Sat, 22 Feb 2025 00:08:53 -0500 (10:38 +0530)
committer	Gyan Doshi <ffmpeg@gyani.pro>	
	Sat, 22 Feb 2025 08:27:53 -0500 (18:57 +0530)
commit	d1ed5c06e3edc5f2b5f3664c80121fa55b0baa95
tree	8870c516275d85f417972d8402ebcce1126d4fdd	tree
parent	392aa9daf2c730465f6aba5ba2b37e02695de9f7	commit | diff
avcodec/libsvtav1: unbreak build with latest svtav1

SVT-AV1 made a change in their public API in 988e930c but without a
version bump or any other accessible marker, thus breaking ffmpeg build
with current versions of SVT-AV1.

They have finally bumped versions a month later, so check added.
```
https://git.ffmpeg.org/gitweb/ffmpeg.git/blob/HEAD:/libavcodec/libsvtav1.c#l438
```
 436     svt_enc->eos_flag = EOS_NOT_REACHED;
 437 
 438 #if SVT_AV1_CHECK_VERSION(3, 0, 0)
 439     svt_ret = svt_av1_enc_init_handle(&svt_enc->svt_handle, &svt_enc->enc_params);
 440 #else
 441     svt_ret = svt_av1_enc_init_handle(&svt_enc->svt_handle, svt_enc, &svt_enc->enc_params);
 442 #endif
 443     if (svt_ret != EB_ErrorNone) {
 444         return svt_print_error(avctx, svt_ret, "Error initializing encoder handle");
 445     }
```

I ran into this issue while building jellyfin-ffmpeg.

regularhunter:
The upstream commit is not present in jellyfin-ffmpeg-7.1.1-1. Still broken

nyanmisaka:
> The upstream commit is not present in jellyfin-ffmpeg-7.1.1-1. Still broken

You may not know that we use patches instead of modifying the source code directly.

[...]

PR #545, opened 2025-03-03
Accomodate SVT-AV1 API change

regularhunter:
SVT-AV1 3.0.0 changed the signature of svt_av1_enc_init_handle. ffmpeg has not released a version including the needed change.

Fixes issue #544


nyanmisaka:
We do not modify the ffmpeg source code directly but use patch.

https://github.com/jellyfin/jellyfin-ffmpeg/tree/jellyfin/debian/patches

We haven't upgraded to SVT-AV1 3.0 yet. When we do it I'll add this as a patch as well.

PR #562, opened 2025-04-28
Fix the transparency issue of ASS subtitle rendering in HWA

nyanmisaka:
Before this change, `vf_subtitles` could only generate subtitle images with premultiplied alpha. However, HW overlay filters usually only support straight alpha (unpremultiplied). This caused some transparency issues, especially around font edges in ASS subtitles.

This change allows `vf_subtitles` to output unpremulplied alpha directly in RGB32/BGR32. And for some HW overlay filters that support premultiplied alpha, or prefer YUVA420P, an `alpha_format` option is added to control whether the overlay input is straight or premultiplied, and process accordingly. The default behavior of HW overlay filters is unchanged.

**Changes**
- Add alpha_format option for OCL/CUDA/VAAPI/RKRGA overlay filters
- Add unpremultiplied alpha support for subtitles filter


gnattu:
Do we need any server side changes to utilize this?

nyanmisaka:
> Do we need any server side changes to utilize this?

Need to check if `overlay_{cuda,opencl}` supports `alpha_format` option, if yes set it to 1 for YUVA420P.

Shadowghost (approved):

PR #564, opened 2025-05-02
avfilter/tonemap_opencl: reduce compute and memory load of 3dlut

gnattu:
This reduces both memory and compute load when applying the 3dlut. Instead of computing all possible tetrahedrons and fetching all eight vertices, it now only fetches four vertices of a single determined tetrahedron. This reduces the memory load to 50% and the ALU load to approximately 33%, while maintaining a branchless flow.

Unfortunately this helps little for 4K dovi reshaping. I've also tried multiple approaches for the reshaping kernel but none of them actually improved performance.

**Changes**

**Issues**

### DIFF
--- a/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
+++ b/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
@@ -593,7 +593,7 @@ Index: FFmpeg/libavfilter/opencl/tonemap.cl
      float j = tone_param;
      float a, b;
  
-@@ -71,202 +85,614 @@ float mobius(float s, float peak) {
+@@ -71,202 +85,615 @@ float mobius(float s, float peak) {
          return s;
  
      a = -j * j * (peak - 1.0f) / (j * j - 2.0f * j + peak);
@@ -1225,7 +1225,8 @@ Index: FFmpeg/libavfilter/opencl/tonemap.cl
 +    unsigned idx101 = base_idx + 1 + LUT_SIZE * LUT_SIZE;
 +    unsigned idx011 = base_idx + LUT_SIZE + LUT_SIZE * LUT_SIZE;
 +
-+    // Although we have a and c as max and min value, we cannot use them in the following selection as float equality comparison is not accurate on GPU.
++    // Although we have a and c as max and min value, we cannot use them in the
++    // following selection as float equality comparison is not accurate on GPU.
 +    unsigned idx0 = select(select(idx001, idx010, (f.y >= f.z && f.y >= f.x)), idx100, (f.x >= f.y && f.x >= f.z));
 +    unsigned idx1 = select(select(idx110, idx101, (f.y <= f.z && f.y <= f.x)), idx011, (f.x <= f.y && f.x <= f.z));
 +
--- a/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
+++ b/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
@@ -593,7 +593,7 @@ Index: FFmpeg/libavfilter/opencl/tonemap.cl
      float j = tone_param;
      float a, b;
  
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
```diff
@@ -1208,40 +1208,36 @@ Index: FFmpeg/libavfilter/opencl/tonemap.cl
...
-+                           (int3)(f.x >= f.y));
++    // Select the index for vertices of the tetrahedron.
++    unsigned idx100 = base_idx + 1;
++    unsigned idx010 = base_idx + LUT_SIZE;
++    unsigned idx110 = base_idx + 1 + LUT_SIZE;
++    unsigned idx001 = base_idx + LUT_SIZE * LUT_SIZE;
++    unsigned idx101 = base_idx + 1 + LUT_SIZE * LUT_SIZE;
++    unsigned idx011 = base_idx + LUT_SIZE + LUT_SIZE * LUT_SIZE;
++
++    // Although we have a and c as max and min value, we cannot use them in the following selection as float equality comparison is not accurate on GPU.
```
```suggestion
+    // Although we have a and c as max and min value, we cannot use them in the
+    // following selection as float equality comparison is not accurate on GPU.
```
Better if it could break a line here.

nyanmisaka (approved):
The shaping is heavily compute-bounded by `mad()` func. Comment out `mix()` or `reshape_poly()` and you can see big speedups on common clips that don't need `reshape_mmr()`.

I once even thought to compute all possible`coeffs[4]` for `IPT.I` in real time on the cpu side and then select on the gpu.

gnattu:
> The shaping is heavily compute-bounded by `mad()` func. Comment out `mix()` or `reshape_poly()` and you can see big speedups on common clips that don't need `reshape_mmr()`.
> 
> I once even thought to compute all possible`coeffs[4]` for `IPT.I` in real time on the cpu side and then select on the gpu.

Well it does not look like we are compute bounded on RK3588. The mix chain is just a selection logic and rewriting them to pure logic assignments which has much ligher ALU load like the following does not help with the performance:

```opencl
float4 coeff01 = (s >= dovi_pivots[0]) ? dovi_coeffs[1] : dovi_coeffs[0];
float4 coeff23 = (s >= dovi_pivots[2]) ? dovi_coeffs[3] : dovi_coeffs[2];
float4 coeff1 = (s >= dovi_pivots[1]) ? coeff23 : coeff01;
float4 coeff45 = (s >= dovi_pivots[4]) ? dovi_coeffs[5] : dovi_coeffs[4];
float4 coeff67 = (s >= dovi_pivots[6]) ? dovi_coeffs[7] : dovi_coeffs[6];
float4 coeff2 = (s >= dovi_pivots[5]) ? coeff67 : coeff45;
coeffs = (s >= dovi_pivots[3]) ? coeff2 : coeff1;
```

If we are indeed computationally limited, the new 3DLUT interpolation should offer a performance boost due to the reduced ALU load. However, it doesn’t seem to be the case. I believe the lengthy process of reading all coefficients from the buffer is the primary factor hindering the performance. The memory load is too heavy for this bandwidth limited GPU when doing 4K tone mapping. 

And if we comment out the whole coefficient pick logic and change the `reshape_poly` to nop (just return `s`), we barely hit 60fps:

```
frame= 3372 fps= 60 q=-0.0 Lsize=  123445KiB time=00:00:56.11 bitrate=18020.7kbits/s speed=0.991x    
```

It now really looks like the 4K60fps dovi tone mapping target is quite unrealistic for RK3588 unless we can find out some way to significantly reduce memory load.

nyanmisaka:
If this is the case, would prefetching the coeffs and pivots and sharing it among the 4 pixels give better results?

But if the on-die register resources are exhausted and global memory has to be accessed again, I am afraid it will not be improved.

gnattu:
> If this is the case, would prefetching the coeffs and pivots and sharing it among the 4 pixels give better results?

I've tried similar things and it actually decreased the performance. The coefficients are 24 float4 in total after all and that is not a small number of variables.

nyanmisaka:
The halved bandwidth requirement of FP16 and vload() vector loading helped. #567

nyanmisaka (approved):

PR #565, opened 2025-05-03
Add V4L2 Request API support

a-andre:
The patches are taken from https://github.com/Kwiboo/FFmpeg/tree/v4l2request-2024-v2

**Changes**

On Rockchip devices the V4L2 Request API provides an alternative to rkmpp. In contrast to rkmpp, the request API works with mainline Linux.

Currently, mpeg2 and h264 hardware acceleration are supported. The [HEVC patch](https://github.com/Kwiboo/FFmpeg/commit/2af4006b7a3506d439eaf725b3d45579740b1026) is not included yet, because of compile issues.

### DIFF
--- a/debian/patches/0089-add-hwdevice-type-for-v4l2-request-api.patch
+++ b/debian/patches/0089-add-hwdevice-type-for-v4l2-request-api.patch
@@ -1,0 +1,443 @@
+commit 278e6cd98fc355dea2f22bbf1c1d7a334f70945a
+Author: Jonas Karlman <jonas@kwiboo.se>
+Date:   Tue Aug 6 09:06:00 2024 +0000
+
+    avutil/hwcontext: Add hwdevice type for V4L2 Request API
+    
+    Add a hwdevice type for V4L2 Request API with transfer_data_from support
+    for AV_PIX_FMT_DRM_PRIME, based on AV_HWDEVICE_TYPE_DRM.
+    
+    AVV4L2RequestDeviceContext.media_fd can be set by the application or a
+    media device path can be supplied when hwdevice is created. When none
+    is supplied it default to -1 and hwaccel will auto-detect a media device
+    with a capable video device.
+    
+    Signed-off-by: Jonas Karlman <jonas@kwiboo.se>
+
+diff --git a/configure b/configure
+index 015a196de3..018e822160 100755
+--- a/configure
++++ b/configure
+@@ -360,6 +360,7 @@ External library support:
+   --enable-rkmpp           enable Rockchip Media Process Platform code [no]
+   --enable-rkrga           enable Rockchip 2D Raster Graphic Acceleration code [no]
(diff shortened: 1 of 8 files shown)


nyanmisaka:
We need end-to-end (decode + RGA + encode) hardware transcoding acceleration. In the best case, this is only 1/3 of the way there. Right now it's more suitable for a single-use-case video player like MPV or Kodi.

https://jellyfin.org/docs/general/post-install/transcoding/hardware-acceleration/rockchip#acceleration-methods

> Until there is a mature transcoding solution in the Linux mainline, users need to use RKMPP.

a-andre:
There seems to be no easy way to use RKMPP with mainline linux. In that case using hardware acceleration for 1/3 would already be an improvement compared to doing all in software.

Mainline linux also supports RGA. But I couldn't figure out if it is supported by FFMPEG or patches exist.

The patches added in this PR might also be taken from https://github.com/jernejsk/FFmpeg/commits/v4l2-request-n7.1/. There, the HEVC patch seems to be fixed and hardware acceleration is also provided for VP8 and VP9.

gnattu:
> There seems to be no easy way to use RKMPP with mainline linux. In that case using hardware acceleration for 1/3 would already be an improvement compared to doing all in software.

We are requiring our users to use the BSP kernel due to the immaturity of hardware support for the mainline Linux kernel. For a SoC with a relatively slow CPU (performance in the N5105 to N100 range), partial acceleration is almost meaningless. Even with the BSP kernel and full GPU acceleration support, the most demanding task, such as Dolby Vision tone mapping, is still bottlenecked and cannot achieve 4K60fps performance. Additionally, the mainline kernel driver does not support OpenCL, which means that this heavy compute task can only be performed on the CPU, defeating the purpose of using the RK3588, which is a low power SoC with decent hardware accelerators. 

We are [deprecating Raspberry Pi V4L2 acceleration](https://jellyfin.org/docs/general/post-install/transcoding/hardware-acceleration/#raspberry-pi-hardware-acceleration-support-deprecation) for almost the same reason. For such a slow SoC, partial acceleration has almost no difference compared to no acceleration. The raspberry Pi V4L2 has a similar 1/3 acceleration situation and it works really really bad.

In my opinion, it’s an excessive task for us to maintain a patchset that doesn’t work well for our primary use case. Moreover, it makes even less sense to maintain a feature that we have to tell our users not to use to achieve optimal performance. 

I understand that people hate BSP kernel and want to use mainline kernels as soon as possible. But in my opinion, properly working hardware is more important than a "correct" kernel version. When the mature support is there, we will make our ffmpeg to work with mainline kernel. But for now, it is not.

a-andre:
> In my opinion, it’s an excessive task for us to maintain a patchset that doesn’t work well for our primary use case. Moreover, it makes even less sense to maintain a feature that we have to tell our users not to use to achieve optimal performance.

I understand. Thanks for the explanation.

It's sad that Rockchip hasn't much interest in mainline linux support.

PR #567, opened 2025-05-04
Add FP16 tradeoff for DOVI reshaping

nyanmisaka:
**Changes**
- Add FP16 tradeoff for DOVI reshaping
- Fix VUI full range flag in RKMPP encoded videos
- Increase default frame pool size for D3D11VA AV1/VP9
- Handle NOPTS and no extradata in RKMPP decoders
- Bump version to 7.1.1-3

TODO:
- [x] Should it be enabled selectively based on tradeoff and fp16_hw_caps?
- [x] Test in various clips to avoid regressions (visible visual degradation)

image

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1,7 +1,7 @@
 ---
 # We just wrap `build` so this is really it
 name: "jellyfin-ffmpeg"
-version: "7.1.1-2"
+version: "7.1.1-3"
 packages:
   - bullseye-amd64
   - bullseye-armhf
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,14 @@
+jellyfin-ffmpeg (7.1.1-3) unstable; urgency=medium
+
+  * avfilter/tonemap_opencl: reduce compute and memory load of 3dlut
+  * avfilter/tonemap_opencl: always use eotf for lms2rgb
+  * Add FP16 tradeoff for DOVI reshaping
+  * Fix VUI full range flag in RKMPP encoded videos
+  * Increase default frame pool size for D3D11VA AV1/VP9
+  * Handle NOPTS and no extradata in RKMPP decoders
+
+ -- nyanmisaka <nst799610810@gmail.com>  Tue, 6 May 2025 21:12:22 +0800
+
 jellyfin-ffmpeg (7.1.1-2) unstable; urgency=medium
 
   * Fix VPP tonemap without having HDR10 metadata
--- a/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
+++ b/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
@@ -499,11 +499,32 @@ Index: FFmpeg/libavfilter/opencl/tonemap.cl
 ===================================================================
 --- FFmpeg.orig/libavfilter/opencl/tonemap.cl
 +++ FFmpeg/libavfilter/opencl/tonemap.cl
-@@ -16,54 +16,68 @@
+@@ -16,54 +16,89 @@
   * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
   */
  
 -#define REFERENCE_WHITE 100.0f
++#ifdef DOVI_RESHAPE
++  #undef typedef_vecs
++  #undef M_ZERO_VEC
++  #define typedef_vecs(base, new) \
++    typedef base     new;    \
++    typedef base##2  new##2; \
++    typedef base##3  new##3; \
++    typedef base##4  new##4; \
++    typedef base##8  new##8; \
++    typedef base##16 new##16;
++  #ifdef DOVI_PERF_FP16
++    #pragma OPENCL EXTENSION cl_khr_fp16 : enable
++    #define M_ZERO_VEC 0.0h
++    typedef_vecs(half,  vec)
++  #else
++    #define M_ZERO_VEC 0.0f
++    typedef_vecs(float, vec)
++  #endif
++  #undef typedef_vecs
++#endif
++
 +#define FLOAT_EPS 1e-6f
 +#define LUT_SIZE 65
 +
@@ -589,7 +610,7 @@ Index: FFmpeg/libavfilter/opencl/tonemap.cl
      float j = tone_param;
      float a, b;
  
-@@ -71,202 +85,612 @@ float mobius(float s, float peak) {
+@@ -71,202 +106,632 @@ float mobius(float s, float peak) {
          return s;
  
      a = -j * j * (peak - 1.0f) / (j * j - 2.0f * j + peak);
@@ -851,7 +872,7 @@ Index: FFmpeg/libavfilter/opencl/tonemap.cl
 -float3 map_one_pixel_rgb(float3 rgb, float peak, float average) {
 -    float sig = max(max(rgb.x, max(rgb.y, rgb.z)), 1e-6f);
 +#ifdef DOVI_RESHAPE
-+float reshape_poly(float s, float4 coeffs) {
++vec reshape_poly(vec s, vec4 coeffs) {
 +    return (coeffs.z * s + coeffs.y) * s + coeffs.x;
 +}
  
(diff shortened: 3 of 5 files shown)


gnattu:
I think this should be enabled selectively as not all hw/driver supports fp16 without issue and having a manual switch allows user to disable this when fp16 is buggy on his hardware.

Also, a lot of desktop GPUs that supports fp16 extension can only do fp16 compute, but the throughput is not higher than fp32, and we end up just losing precisions on such GPU. Maybe only default fp16 to on only on ARM Mali GPUs is the safest option.

nyanmisaka:
> I think this should be enabled selectively as not all hw/driver supports fp16 without issue and having a manual switch allows user to disable this when fp16 is buggy on his hardware.
> 
> Also, a lot of desktop GPUs that supports fp16 extension can only do fp16 compute, but the throughput is not higher than fp32, and we end up just losing precisions on such GPU. Maybe only default fp16 to on only on ARM Mali GPUs is the safest option.

UHD6xx also needs this rescue. IIRC it also behaves badly with reshaping. As far as our supported hardware is concerned, the latest drivers for Intel/AMD/ARM should be safe, unless one is forced to use ancient drivers, which may contain bugs. As for NVIDIA, their fp16 extension is experimental and lacks some built-in half functions. Plus with 1:64 fp16 perf on old gen and 1:1 fp16 on current gen, there is no reason to use it.

gnattu:
I've tested this on various samples and unfortunately 4K60fps still cannot be guaranteed as some picky inputs could reduce the performance to ~55fps. But it is safe to say that we can have 2x24fps dovi 4k tone mapping now.

> As far as our supported hardware is concerned, the latest drivers for Intel/AMD/ARM should be safe, unless one is forced to use ancient drivers, which may contain bugs.

Or we can make it simpler, only reduce precision on platforms where tradeoff mode is required and the driver reports support for fp16. We still need a full-precision fp32 pipeline as a fallback, anyway, because a significant number of users have old hardware or use some kind of NAS where the OS is limited in some way.

nyanmisaka:
> I've tested this on various samples and unfortunately 4K60fps still cannot be guaranteed as some picky inputs could reduce the performance to ~55fps. But it is safe to say that we can have 2x24fps dovi 4k tone mapping now.

The remaining variables should be related to chip lottery (rockchip pvtm), temperature, video bitrate and complexity in dovi metadata.
```
$ sudo dmesg | grep pvtm
[   16.961764] rockchip-pvtm fda40000.pvtm: pvtm@0 probed
[   16.961818] rockchip-pvtm fda50000.pvtm: pvtm@1 probed
[   16.961866] rockchip-pvtm fda60000.pvtm: pvtm@2 probed
[   16.961912] rockchip-pvtm fdaf0000.pvtm: pvtm@3 probed
[   16.961960] rockchip-pvtm fdb30000.pvtm: pvtm@4 probed
[   17.099308] cpu cpu0: pvtm=1557
[   17.099560] cpu cpu0: pvtm-volt-sel=6
[   17.108896] cpu cpu4: pvtm=1782
[   17.113173] cpu cpu4: pvtm-volt-sel=7
[   17.123270] cpu cpu6: pvtm=1788
[   17.127543] cpu cpu6: pvtm-volt-sel=7
[   17.407526] mali fb000000.gpu: pvtm=938
[   17.407670] mali fb000000.gpu: pvtm-volt-sel=5
[   17.437814] RKNPU fdab0000.npu: pvtm=953
[   17.443103] RKNPU fdab0000.npu: pvtm-volt-sel=5
```

> Or we can make it simpler, only reduce precision on platforms where tradeoff mode is required and the driver reports support for fp16. We still need a full-precision fp32 pipeline as a fallback, anyway, because a significant number of users have old hardware or use some kind of NAS where the OS is limited in some way.

I ended up choosing this one as well.

gnattu:
I noticed that the tonemapping result from RK3588 is different from the one on my MacBook and is also different from the Intel iGPU. I checked if that is due to fp16 but it is not. Is the dolby vision metadata really correctly handled on the RK3588 side? The low light area is significantly darker than other platforms on RK3588. It is not due to the EOTF being omitted either, as I tested with and without those on other platforms as well.

### RK3588 `tradeoff=0`
Screenshot 2025-05-06 at 09 05 30

### M4 Max `tradeoff=0`
Screenshot 2025-05-06 at 09 04 37

If the dovi metadata is correctly handled, then it probably be cause by something worse like the rkrga is modifying the base layer (I hope it does not).

nyanmisaka:
The dovi rpu parsing is done by ffmpeg, I didn't use the dovi parser built into MPP. All I did was make sure the PTS was in sync.

nyanmisaka:
You can disable decoder afbc and hwdownload to nv15, and convert to 420p10le and upload to opencl. See how it performs.

gnattu:
> You can disable decoder afbc and hwdownload to nv15, and convert to 420p10le and upload to opencl. See how it performs.

Unfortunately, removing `-afbc rga` and doing manual `hwdownload,format=nv15,format=p010,hwupload=derive_device=opencl` does not fix this. So the problem looks like is straight from the decoder frame. Does Rockchip decoder handles full range signals correctly? This could be the problem as dovi p5 always uses full range.

nyanmisaka:
Have you tried setting tonemap_opencl to output limited range? It could also be that the MPP encoder doesn't handle the bt709 full range correctly. I'm not sure because I haven't tested this kind of SDR sample.

gnattu:
> Have you tried setting tonemap_opencl to output limited range? It could also be that the MPP encoder doesn't handle the bt709 full range correctly. I'm not sure because I haven't tested this kind of SDR sample.

This fixed it:
Screenshot 2025-05-06 at 09 42 49

So rkmpp encoder cannot do full range yuv encoding, at least not in bt709. We need to modify the server code to always use limited range and ignore whatever user sets, and maybe hide that option when rkmpp is selected.

nyanmisaka:
There are some yuv and rgb coeffs predefined in MPP runtime, which may contain errors. It shouldn’t be too difficult to fix it there.

gnattu:
I pushed two more changes in #564, probably need to merge that first and rebase this with that. If that is too hard to work with due to the patch system, it is also fine to just incorporate the changes here.

nyanmisaka:
Nice, so it looks like there is no need to omit the PQ eotf. Or you can simply raise an issue in the MPP repo using the pics above. Someone should take a look at it soon.

https://github.com/rockchip-linux/mpp/blob/ff3ae5c01044bab536a520a3c97f1ec85cb4f78b/mpp/hal/rkenc/h265e/hal_h265e_vepu580.c#L2052

nyanmisaka:
A bonus question, would switching LUT3D to `cl_half` also squeeze out any extra performance?

Shadowghost (approved):

gnattu:
> A bonus question, would switching LUT3D to cl_half also squeeze out any extra performance?

It could, but the actual speedup is yet to be determined and likely minimal. The overhead of the 3dlut applying logic is already quite lightweight with only four 128-bit memory reads, ~10 vector operations, and ~20 scalar operations plus moves. If memory bandwidth remains a bottleneck, this may not lead to noticeable performance improvements. 

What I do think would be helpful is to make AFBC work with p010. Currently rga cannot output p010 frames with afbc.

nyanmisaka:
The dev of libMali said importing AFBC surface in OpenCL runtime is supported, but when I queried it through OCL interface, I got no support. I will try to write to him.

```
https://github.com/KhronosGroup/OpenCL-Docs/pull/1019
```

Considering that the extension is still a draft, and the use cases are few and far between, I'm not sure if they've actually implemented this.

gnattu:
> The dev of libMali said importing AFBC surface in OpenCL runtime is supported, but when I queried it through OCL interface, I got no support.

If it is missing in the common releases of libmali then it means most of our users won't have access to it.

nyanmisaka:
> > The dev of libMali said importing AFBC surface in OpenCL runtime is supported, but when I queried it through OCL interface, I got no support.
> 
> If it is missing in the common releases of libmali then it means most of our users won't have access to it.

So I wouldn't bet on it. If it is supported in the future, it should dynamically detect OpenCL capabilities and decide whether to enable it in RGA. On the other hand, Mesa panthor/OGL can support it, so it should not be a hardware limitation.

https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/32691
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/34510

PR #569, opened 2025-05-07
avfilter/tonemapx: use cpu friendly conditional assignments

gnattu:
The orignial logic ported from the GPU shader uses (fused) multiply and add to compute the dovi coefficients. But we are not actually blending the values, and we are selecting all lanes based on one boolean condition, a simple conditional assignment is much faster on CPU.

This commit also uses casting instead of extraction to get the low 128b for the avx code path. Casting is free but extraction is an instruction with 3-cycle latency.

#### M4 Max with NEON:

Before: 

`frame= 3372 fps= 81 q=-1.0 Lsize=  281226KiB time=00:00:56.16 bitrate=41017.2kbits/s speed=1.35x`

After:

`frame= 3372 fps= 93 q=-1.0 Lsize=  281212KiB time=00:00:56.16 bitrate=41015.2kbits/s speed=1.55x`

About 15% faster

#### Intel Core i9-12900 with AVX:

Before:

`frame= 3372 fps= 64 q=-1.0 Lsize=  280936KiB time=00:00:56.16 bitrate=40975.0kbits/s speed=1.07x`

After:

`frame= 3372 fps= 69 q=-1.0 Lsize=  280926KiB time=00:00:56.16 bitrate=40973.5kbits/s speed=1.14x`

About 8% faster

**Changes**

**Issues**

### DIFF
--- a/debian/patches/0057-add-simd-optimized-tonemapx-filter.patch
+++ b/debian/patches/0057-add-simd-optimized-tonemapx-filter.patch
@@ -94,7 +94,7 @@ Index: FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/aarch64/vf_tonemapx_intrin_neon.c
-@@ -0,0 +1,2366 @@
+@@ -0,0 +1,2360 @@
 +/*
 + * Copyright (c) 2024 Gnattu OC <gnattuoc@me.com>
 + *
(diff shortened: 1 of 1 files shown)


Shadowghost (approved):

nyanmisaka (approved):

nyanmisaka:
Somehow the MIX() seems to be slightly faster on the GPU.

gnattu:
> Somehow the MIX() seems to be slightly faster on the GPU.

Yes, that is the reason libplacebo uses mix in the shader. A lot of GPUs have lower throughputs for its predicate instructions and FMA ops just have higher throughput, on CPU it is usually the other way around, FMA is extremely heavier than predicates. This is the result of design differences between CPU and GPU.

PR #573, opened 2025-05-09
Add intel-opencl-icd to control/Suggests

nyanmisaka:
**Changes**
- Add `intel-opencl-icd{,-legacy}` to control/Suggests

  So that users can install Intel OpenCL runtime more conveniently by adding `--install-suggests` to `apt-get`.


Shadowghost (approved):
Are those even in the main repos?

nyanmisaka:
> Are those even in the main repos?

They are installed directly from the debian/ubuntu repo.

Some users don't use our docker images, and don't read the HWA documentation. This is for them.

PR #577, opened 2025-05-14
avfilter/tonemap_opencl: add scalar reshaping for qcom gpu

gnattu:
The Qualcomm's own OpenCL driver has very inconsistent performance that a general vectorization optimization would lead to performance regressions. Adding back the scalar code path for this platform as this is the only way that I found would make its OpenCL performance to not suffer.

Besides the Qualcomm's official driver, there is also a CLonD3D12 compatibility layer downloadable from Microsoft Store. This compatibility layer is slightly slower than the official driver when using the "optimized for qualcomm" kernel, but this driver has more predictable performance and behaves more like other GPUs. The vectorized kernel actually helps with the performance, but still slightly slower.

**Changes**

**Issues**

### DIFF
--- a/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
+++ b/debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
@@ -2159,7 +2159,7 @@ Index: FFmpeg/libavfilter/vf_tonemap_opencl.c
 +                }
 +                av_free(device_name);
 +            }
-+        } else if (device_vendor_id == 0x5143 || device_vendor_id == ('Q' | 'C' << 8 | 'O' << 16 | 'M' << 24)) {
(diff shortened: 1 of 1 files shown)


debian/patches/0007-add-bt2390-eetf-and-code-refactor-to-opencl-tonemap.patch
```diff
@@ -2048,6 +2159,9 @@ Index: FFmpeg/libavfilter/vf_tonemap_opencl.c
 +                }
 +                av_free(device_name);
 +            }
++        } else if (device_vendor_id == 0x5143 || device_vendor_id == ('Q' | 'C' << 8 | 'O' << 16 | 'M' << 24)) {
```
```suggestion
+        } else if (device_vendor_id == 0x5143 || device_vendor_id == MKTAG('Q', 'C', 'O', 'M')) {
```

nyanmisaka (approved):

Shadowghost (approved):

PR #582, opened 2025-05-23
Apply DOVI L1 metadata when available

nyanmisaka:
**Changes**
- Apply DOVI L1 metadata when available (non-3DLUT paths)
- Build libunibreak for libass so that SRT subs wrap correctly
- Allow QSV Enc to bail out from unsupported low-power mode
  (Tired of seeing more troubleshooting cases about this)


builder/scripts.d/45-libunibreak.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+SCRIPT_REPO="https://github.com/adah1972/libunibreak.git"
+SCRIPT_COMMIT="32dc4fabf20176aeaa63c4880cc53cc123e53261"
+
+ffbuild_enabled() {
+    return 0
```
gnattu, line 7: libunibreak is already part of the macOS system image dependency: https://github.com/jellyfin/jellyfin-ffmpeg/blob/jellyfin/builder/images/macos/06-libunibreak.sh

If we want to build it here we should also remove it from the system dependency

gnattu:
~~Why the L1 metadata for brightness is not applied for trade off mode? It does not introduce extra compute cost from my observation and it only changes the peak value being passed into the kernel.~~

NVM, it requires the LUT to be rebuilt for every scene changes.

debian/patches/0004-add-cuda-tonemap-impl.patch
```diff
@@ -104,29 +104,34 @@ Index: FFmpeg/libavfilter/colorspace.c
...
++    if (!l0_only && (ext = av_dovi_find_level(data, 1)))
++        peak = ext->l1.max_pq / 4095.0f;
 +    if (!peak)
 +        return peak;
 +
 +    peak = powf(peak, 1.0f / ST2084_M2);
 +    peak = fmaxf(peak - ST2084_C1, 0.0f) / (ST2084_C2 - ST2084_C3 * peak);
 +    peak = powf(peak, 1.0f / ST2084_M1);
-+    peak *= 100.0f;
++    peak *= (ST2084_MAX_LUMINANCE / REFERENCE_WHITE_ALT);
```
gnattu: I'm afraid this is not entirely correct. The computed signal peak is now too low. The libav's internal peak has nothing to do with our 203nit reference white but is hardcoded as 1 = 100nit, so as all our shaders.

debian/patches/0050-add-vf-tonemap-videotoolbox-filter.patch
```diff
@@ -1869,11 +1870,12 @@ Index: FFmpeg/libavfilter/vf_tonemap_videotoolbox.m
 +    if (!ctx->peak) {
 +        if (dovi_sd) {
 +            const AVDOVIMetadata *metadata = (AVDOVIMetadata *) dovi_sd->data;
-+            ctx->peak = ff_determine_dovi_signal_peak(metadata);
-+        } else {
-+            ctx->peak = ff_determine_signal_peak(input);
++            ctx->src_peak = ff_determine_dovi_signal_peak(metadata, 0);
```
gnattu, line 1871: ~~If assigning this way we probably cannot just use this method because the scale returned by this function is not the same as we are expecting within our kernel. In our kernel, 1 = 10 nit, but it seems like this internal ffmpeg method assumes 1 = 100nit~~

nvm, we divide by 10 for human inputs. I mixed those.

nyanmisaka: Put it into a draft then. I'll look into rescaling properly later.

gnattu: Current way is okay, if the returned value is within the same scale(1=100nit)

debian/patches/0004-add-cuda-tonemap-impl.patch
gnattu: I *think* keep the old 100.0f is the correct thing to do for now. The old 100.0f has the same scale with the non-dv variants where the peak is multiplied by 10000 / 100.

nyanmisaka: In previous commits, the reasons for low brightness are:

1. For the DOVI + LUT case, the peak uses the wrong L1 max of the first frame instead of the L0 max. The first frame may be a black frame, so the brightness is extremely low. This is now fixed.

2. For HDR10, the peak in the static metadata is unreliable and may even be lower than or equal to the SDR peak `1.0f`. This causes issues in the tonemap algorithms. A sanity check is needed to make it fall back to a valid value such as `10.0f` (1000 nits). This is now fixed.

Regarding whether to introduce `203.0f` when calculating DOVI peak, I refer to libplacebo:
`(max_pq_from_dv_metadata / 4095.0f)` corresponds to [`PL_HDR_PQ`](https://github.com/haasn/libplacebo/blob/3188549fba13bbdf3a5a98de2a38c2e71f04e21e/src/include/libplacebo/colorspace.h#L363). In our case, the value is rescaled from `PL_HDR_PQ` to `PL_HDR_NORM` through [`pl_hdr_rescale()`](https://github.com/haasn/libplacebo/blob/3188549fba13bbdf3a5a98de2a38c2e71f04e21e/src/colorspace.c#L367). So the `eotf_st2084_common()` and `/= 203.0f` are used here.

nyanmisaka: I may be wrong because PL_HDR_NORM range seems to be [0, 1] `PL_HDR_NORM = 0, // 0.0 is absolute black, 1.0 is PL_COLOR_SDR_WHITE`. And we need the peak to be much larger than this. `/100.0f` may be more correct here, but I can't find its direct mapping in libplacebo.

gnattu: Well that is due to the difference between libplacebo and ffmpeg. For libplacebo, the signal value of 1.0 in its shader is already 203nit but for our shader a signal value of 1.0 implies 100nit due to historical reasons. The ffmpeg native peak derive function assumes the 100nit white scale: https://github.com/FFmpeg/FFmpeg/blob/670089304ac856f01591c88bbb7c115d423935da/libavfilter/colorspace.c#L153-L176

Which means our kernel currently behaves a bit differently when auto brightness for hdr10 and dolby vision inputs and even with manual peak value. If we want to rescale the input peak to the level of reference sdr white of 203, I *think* we have to do the similar rescaling everywhere but that also means we are effectively changing our default behavior, and that is what I'm a bit afraid of, although that might be technically more correctly to do as we no longer use the 100nit as reference white for PQ inputs now.

gnattu:
So this is my brief summary of current situation an my plan to make the current behavior more consistent:

ffmpeg uses 100 nits as reference white, our code uses 203 nits. This mismatch causes normalization inconsistencies for signal/luminance and makes tone mapping curves behave differently.

For most of our tone mapper functions, we scale the curve to the input range of [0, input signal peak] so that the input value of the peak outputs exactly 1.0. However, this current inconsistency results in a slightly flatter slope for such tone mappers because the curve for 100 nit has a higher normalized peak value compared to the curve for 203 nit.

Our HLG normalization is also off, since we’re not correctly mapping its reference white and output range.

For HLG signals, “reference white” is defined as a signal value of 0.75. The EOTF for HLG gives approximately 3.179 at this point. The HLG EOTF has a maximum output value of 12.0 (when input is exactly 1.0). To correctly rescale the HLG signal to our linear range as PQ, we need to multiply the output of the OOTF for 1000nit(the `ootf_1_2` we are currently using) by 12, then divide by the 75% value (approximately 3.179). This will make the HLG signal behave similarly to the PQ one, which multiplies by 10000 nit (PQ’s maximum luminance) and divides by 203 (reference white).

For our convenience, we don't touch the already there ffmpeg function. Instead, every time we update/receive the signal peak (from ffmpeg derivation method or user), convert that value to absolute nits, max out at PQ max 10000, then divide by 203 nits to match our kernel’s expectations. This would also require the dovi derivation method here to use the ffmpeg scale (100nit reference white) if we want to share the same logic for all peak sources.

For HLG, we fix it in our kernel and that is straightforward. We just change the mysterious `5.0f` multiplier to `12/3.179f` and we are done.

debian/patches/0004-add-cuda-tonemap-impl.patch
nyanmisaka: Indeed. It looks like mpv/libplacebo has switched to 203.0f internally since the introduction of the bt2390 tonemapper.

https://github.com/mpv-player/mpv/commit/ef6bc8504a945eb6492b8ed46fd5a1afaaf32182

msys2/PKGBUILD/20-mingw-w64-libxml2/PKGBUILD
```diff
@@ -27,15 +27,11 @@ install=${_realname}-${MSYSTEM}.install
 source=("https://download.gnome.org/sources/libxml2/${pkgver%.*}/${_realname}-${pkgver}.tar.xz"
         https://www.w3.org/XML/Test/xmlts20130923.tar.gz
         0027-decoding-segfault.patch
-        0029-xml2-config-win-paths.patch
-        0030-pkgconfig-add-Cflags-private.patch
-        0031-apply-msvc-relocation.patch)
-sha256sums=('c3d8c0c34aa39098f66576fe51969db12a5100b956233dc56506f7a8679be995'
+	0029-xml2-config-win-paths.patch)
```
gnattu, line 30: I'm a bit unconfident about dropping upstream(msys2) patches and upgrading to a version later than the upstream. Does the patches being dropped affects our use cases?

nyanmisaka: If I read it correctly, 0030 and 0031 have been fixed in libxml2 2.13 and 2.14.
https://github.com/GNOME/libxml2/commit/c106455c2571c20a0c66942db29247490ac626e9
https://github.com/GNOME/libxml2/commit/a9e73f11c0e08531e4eefe9c4e3983f2b549f700

Shadowghost (approved):

gnattu (approved):

gnattu (approved):

PR #590, opened 2025-06-08
Add NV15 format input support for certain OpenCL filters

nyanmisaka:
**Changes**
- Add NV15 format input support for certain OpenCL filters
  - DRM_PRIME (NV15) -> OpenCL sharing
  - NV15 input support for `vf_{scale,tonemap}_opencl` filters

GPGPU have no native 10bit type support. De-compact is achieved through bit shifting. It's intended to be used as a fallback path and for testing. It runs quite fast on the Mali-G610.

**Issues**
- Some chips with RGA versions that do not support P010 output can use this path
  - e.g. RK3576 [Mali-G52 MC3](https://www.rock-ap.co.kr/uploads/RK3576%20Brief%20Specification.pdf) has a 145 GFOPS FP32 but 4k60+ encoders (Unsure how our kernel will perform...)

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1,7 +1,7 @@
 ---
 # We just wrap `build` so this is really it
 name: "jellyfin-ffmpeg"
-version: "7.1.1-5"
+version: "7.1.1-6"
 packages:
   - bullseye-amd64
   - bullseye-arm64
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+jellyfin-ffmpeg (7.1.1-6) unstable; urgency=medium
+
+  * lavc/videotoolbox: don't return external error for invalid vt frame
+  * Add NV15 format input support for certain OpenCL filters
+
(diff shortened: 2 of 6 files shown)


Shadowghost (approved):

gnattu (approved):
Doing de-interleaving with GPU shaders might have slightly precision issues, and with tonemap filter we might have some scenes(more likely with high luminance) that could make the color look a bit off if the precision loss is high enough. But if performance gain from this is big enough this can be treated as a tradeoff.

nyanmisaka:
> Doing de-interleaving with GPU shaders might have slightly precision issues, and with tonemap filter we might have some scenes(more likely with high luminance) that could make the color look a bit off if the precision loss is high enough. But if performance gain from this is big enough this can be treated as a tradeoff.

Updated now to make it bit exact as p010. No more precision issues.

gnattu (approved):

PR #594, opened 2025-06-14
Add windows arm builder, disable gcc build for windows

gnattu:
**Changes**

This adds windows arm64 builder based on msys2 and its clang arm64 toolchain. Due to the toolchain constraints, the arm64 build has to be run on Windows arm64 runners, and unfortunately, some of the tools has to be run under x86 emulation, which could make the build taking longer than expected.

This PR also disabled the gcc runner for now. Our Clang build has been included by default in our windows releases for the entire 10.10 cycle and we did not receive any compatibility reports due to the compiler yet. The gcc build scripts is now in achieved mode and is only being used for downstream packagers/end users to cross build it on Linux if needed.

**Issues**

### DIFF
--- a/.github/workflows/_meta_win_clang_portable.yaml
+++ b/.github/workflows/_meta_win_clang_portable.yaml
@@ -16,13 +16,25 @@ on:
 jobs:
   build:
     name: 'Build Portable FFmpeg'
-    runs-on: windows-latest
+    runs-on: ${{ matrix.os.runner }}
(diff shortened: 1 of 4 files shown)


nyanmisaka:
Predictably some external users will start complaining after tagging the next version.

Shadowghost (approved):

nyanmisaka (approved):

PR #618, opened 2025-09-05
Add nproc alias for macOS builder

gnattu:
**Changes**

nproc is not a macOS built-in command and the closest equivalent is `sysctl -n hw.logicalcpu`. The GitHub runner used to provide the nproc alias but it seems like this is no longer the case. Make a script to redirect nproc instead.

**Issues**

### DIFF
--- a/builder/images/macos/00-dep.sh
+++ b/builder/images/macos/00-dep.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
 
 ffbuild_macbase() {
-  brew install wget subversion mercurial autoconf automake cmake meson ninja pkg-config coreutils gcc make python-setuptools pcre2 libtool gnu-sed gnu-tar nasm quilt texinfo coreutils
+  # https://github.com/actions/runner-images/issues/12912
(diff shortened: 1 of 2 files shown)


Shadowghost (approved):

gnattu:
It looks like the zlib build has to be fixed separately and missing nrpoc alias is only one problem

gnattu:
Alright I found the core issue, microsoft is doing stupid things to its macOS image:
https://github.com/actions/runner-images/issues/12912

Full story:

CMake 4.x update broke backwards compatibility (we had already prepared for this change in #568). However, many users had not, and the macOS github runner image maintainer decided to pin the CMake package to version 3.x:https://github.com/actions/runner-images/pull/12791

After this change was rolled out, our CI was completely broken due to the pinned CMake version. We are not the only ones affected; see the linked issue for other affected users.

This is a very unfortunate situation. The intention behind pinning CMake—to prevent CI breaks—had the direct opposite effect on another group of users including us.

nyanmisaka (approved):

Bond-009 (approved):

PR #639, opened 2025-11-19
Disable mold linker to save build time and space

nyanmisaka:
**Changes**
- Disable mold linker to save build time and space

### DIFF
--- a/builder/images/base-linux64/ct-ng-config
+++ b/builder/images/base-linux64/ct-ng-config
@@ -747,35 +747,7 @@ CT_CC_LANG_OTHERS=""
 #
 # BFD enabled in binutils
 #
-CT_LINKER_MOLD=y
-CT_LINKER_MOLD_PKG_KSYM="MOLD"
-CT_MOLD_DIR_NAME="mold"
-CT_MOLD_PKG_NAME="mold"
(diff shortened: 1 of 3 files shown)


Gauvino:
Do you have any comparaison with before and after ?

nyanmisaka:
No, this is an attempt to fix the recent "no space left on the device" issue. But it's not enough.

lawrencecurtis:
I found using jlumbroso/free-disk-space@main before build portable helped a lot

satmandu:
For what it's worth, we use mold linker in our chromebrew package builds using GitHub Actions, because `mold` is very, very fast, and the `mold` dev is very responsive. Maybe give @ru314 a ping if there's a mold issue with the builds?

@lawrencecurtis Thanks for the suggestion to try [jlumbroso/free-disk-space](https://github.com/jlumbroso/free-disk-space)!

Oh, and I guess there's a more recently updated fork of that: https://github.com/BRAINSia/free-disk-space

nyanmisaka:
Thanks for the reminder. The previously introduced mold linker did increase the time required to build the toolchain and increased the risk of OOM errors in multi-core, low-memory environments. However, disabling it did not help with disk space.

Will use `free-disk-space` in #641

PR #678, opened 2026-04-02
update license file with additional library license groups

jumoog:
**Changes**
Added documentation for 14 missing external dependencies to LICENSE.md that are actively built and linked in jellyfin-ffmpeg but were not previously documented. This includes chromaprint (audio fingerprinting), AV1 codecs (dav1d, svtav1), subtitle support (libass), Blu-ray support (libbluray), and platform-specific hardware acceleration libraries. Dependencies are organized by license type (GPL v2, LGPL v2.1+, BSD 2-Clause, ISC, MIT) to clarify their compatibility with FFmpeg's licensing.

**Issues**
Completes the license documentation for all actively built external libraries. Addresses the missing chromaprint and other undocumented dependencies that should be listed in the license file.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -95,14 +95,31 @@ The following libraries are under GPL version 2:
 - libxavs
 - libxavs2
 - libxvid
+- zvbi
+- openmpt
 
 When combining them with FFmpeg, FFmpeg needs to be licensed as GPL as well by
 passing `--enable-gpl` to configure.
+
+The following libraries are under LGPL version 2.1 or later:
+- chromaprint
+- libbluray
 
 The following libraries are under LGPL version 3:
 - gmp
 - libaribb24
 - liblensfun
+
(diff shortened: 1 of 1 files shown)


nyanmisaka:
This file is part of the upstream project. We do not modify it (https://github.com/jellyfin/jellyfin-ffmpeg/commits/jellyfin/LICENSE.md). We only maintain patches to the source code. The `configure` file already lists the components that are compatible and incompatible with the GPL (https://github.com/jellyfin/jellyfin-ffmpeg/blob/jellyfin/configure#L1841-L2069). I believe it is simply that no one bothers to update this file unless an issue arises.

So I think cosmetic changes should be submitted to the upstream repository (https://code.ffmpeg.org/). This file will be updated when we pull a new release from upstream.

jumoog:
ok

PR #681, opened 2026-04-04
ffprobe: avoid runaway only_first_vframe probing

### LINKED ISSUE #16549: v10.11.7 and ffprobe ram usage
Since I have to systematically recreate my libraries from scratch, as mentioned in https://github.com/jellyfin/jellyfin/issues/15518, I've realized with the release of version 10.11.7 that I can no longer even rescan all the movies in the folder.

I investigated and eventually realized that 8GB of memory on the server is no longer sufficient to scan a 4TB disk of movies.
So I run a 'refresh' of the library and monitor what's happening in a window using the 'htop' command. I see the memory usage gradually increase, reaching 100% RAM usage, then the swap space consumption increases, but it's only 1GB, so it's saturated in 3 seconds. Then the system seems to freeze, or at least very slow, and then Jellyfin restarts on its own.

The server is a VM, so I allocated 16GB, but it made no difference. I also allocated 24GB, and it's the same.

However, there's a setting in the parameters where you can adjust the number of threads. I hadn't configured anything, so I set both fields to 2. With this, I only see two simultaneous ffprobe processes, and RAM consumption reaches about 15GB before dropping, then rising again, and so on.

A least there are no more crashes, so it's still usable. I haven't tried to determine if it actually takes significantly longer than before.

In any case, with version 10.11.6, 8GB was sufficient on the VM.

There's a regression somewhere.

gnattu:
`only_first_vframe` originally waited for the first decoded frame from every video stream. That breaks on files with multiple video streams where one target stream never reaches "first frame seen" in this path.

For example, a Matroska file with one real HEVC video stream and two attached-picture streams and one of them is PNG. With `-threads 1`, all three streams were marked finished quickly as `avcodec_receive_frame()` returns a frame for all three streams. With `-threads 0`, the attached PNG stream was opened with auto threading, its only packet was consumed, but `avcodec_receive_frame()` returned `EAGAIN` and no frame was emitted with multi thread decoding. Since that stream had no second packet, it never became finished. ffprobe then kept demuxing the rest of the file while waiting for all target video streams to finish, which could read essentially to EOF and appear to consume unbounded memory.

Fix this in two ways:
- exclude attached-picture streams from the `only_first_vframe` target set, so the mode only waits for real video streams
- change the safety fallback from counting packets for unresolved target streams to counting overall scanned packets, so even unknown future cases still terminate instead of scanning forever

This preserves the intended "first real video frame" behavior while making the mode safe for problematic multi-stream inputs.

**Issues**

Fixes https://github.com/jellyfin/jellyfin/issues/16549

### DIFF
--- a/debian/patches/0079-add-first-vframe-only-to-ffprobe.patch
+++ b/debian/patches/0079-add-first-vframe-only-to-ffprobe.patch
@@ -9,43 +9,54 @@ Index: FFmpeg/fftools/ffprobe.c
 +static int only_show_first_video_frame = 0;
 +
 +#define ONLY_FIRST_VFRAME_MAX_PACKETS_PER_STREAM 4096
-+#define ONLY_FIRST_VFRAME_MAX_PACKETS_TOTAL     65536
++#define ONLY_FIRST_VFRAME_MAX_PACKETS_SCANNED   65536
 +#define ONLY_FIRST_VFRAME_MAX_DECODE_ITERATIONS 4096
 +
  static char *output_format;
  static char *stream_specifier;
  static char *show_data_hash;
-@@ -3108,9 +3114,12 @@ static int read_interval_packets(WriterC
+@@ -3108,9 +3114,13 @@ static int read_interval_packets(WriterC
      AVFormatContext *fmt_ctx = ifile->fmt_ctx;
      AVPacket *pkt = NULL;
      AVFrame *frame = NULL;
 -    int ret = 0, i = 0, frame_count = 0;
 +    int ret = 0, i = 0, frame_count = 0, nb_video_streams = 0;
-+    int nb_finished_video_streams = 0, total_video_packets = 0;
++    int nb_finished_video_streams = 0, total_packets_scanned = 0;
      int64_t start = -INT64_MAX, end = interval->end;
      int has_start = 0, has_end = interval->has_end && !interval->end_is_offset;
 +    unsigned char *finished_video_stream = NULL;
++    unsigned char *target_video_stream = NULL;
 +    unsigned int *video_packets = NULL;
  
      av_log(NULL, AV_LOG_VERBOSE, "Processing read interval ");
      log_read_interval(interval, NULL, AV_LOG_VERBOSE);
-@@ -3149,6 +3158,25 @@ static int read_interval_packets(WriterC
+@@ -3149,6 +3159,35 @@ static int read_interval_packets(WriterC
          ret = AVERROR(ENOMEM);
          goto end;
      }
 +
 +    if (only_show_first_video_frame) {
-+        int si = 0;
++        int si = 0, have_non_attached_video = 0;
 +        finished_video_stream = av_calloc(ifile->nb_streams, sizeof(*finished_video_stream));
++        target_video_stream = av_calloc(ifile->nb_streams, sizeof(*target_video_stream));
 +        video_packets = av_calloc(ifile->nb_streams, sizeof(*video_packets));
-+        if (!finished_video_stream || !video_packets) {
++        if (!finished_video_stream || !target_video_stream || !video_packets) {
 +            ret = AVERROR(ENOMEM);
 +            goto end;
 +        }
 +        for (si = 0; si < ifile->nb_streams; si ++) {
 +            AVCodecParameters *par = ifile->streams[si].st->codecpar;
-+            if (par->codec_type == AVMEDIA_TYPE_VIDEO) {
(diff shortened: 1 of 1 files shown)


nyanmisaka (approved):

Shadowghost (approved):

nyanmisaka:
In the long run, is it possible to convert it to ffmpeg usage `-vframes/-frames:v`, and then submit it upstream, just to reduce the maintenance burden and avoid similar issues?
```
-frames[:stream_specifier] framecount (output,per-stream)
Stop writing to the stream after framecount frames.
```

gnattu:
> In the long run, is it possible to convert it to ffmpeg usage -vframes/-frames:v, and then submit it upstream, just to reduce the maintenance burden and avoid similar issues?

I'm afraid it would be even harder because we need to cover way more use cases to make it a general flag and more importantly the most obvious usage of `-frames: v` does not help with current situation where we 1. don't care about still images in the container 2. some image demuxer needs special handling to even work with auto threading.

Even with our current limited use case, I have to ask our users to provide samples to figure out what could be wrong and the edge cases we have to handle for upstreaming might be much more than that.

nyanmisaka:
> > In the long run, is it possible to convert it to ffmpeg usage -vframes/-frames:v, and then submit it upstream, just to reduce the maintenance burden and avoid similar issues?
> 
> I'm afraid it would be even harder because we need to cover way more use cases to make it a general flag and more importantly the most obvious usage of `-frames: v` does not help with current situation where we 1. don't care about still images in the container 2. some image demuxer needs special handling to even work with auto threading.
> 
> Even with our current limited use case, I have to ask our users to provide samples to figure out what could be wrong and the edge cases we have to handle for upstreaming might be much more than that.

That makes sense. Currently, as long as the upstream doesn't refactor ffprobe, maintaining patches is relatively easy compared to extending to image and audio frames. The key issues are the scarcity of sample files that can trigger mem leaks and a lack of understanding of ffprobe's internal architecture.

gnattu:
@nyanmisaka Maybe we should release a hotfix version for this patch?

nyanmisaka:
> @nyanmisaka Maybe we should release a hotfix version for this patch?

Yes, you can. However, I saw in that ticket that a few individuals reported that this fix didn't work for them. But most of the feedback was positive.

gnattu:
> a few individuals reported that this fix didn't work for them

A lot of them may not even be using the artifact builds and are using other workarounds.

ClementToche:
> > a few individuals reported that this fix didn't work for them
> 
> A lot of them may not even be using the artifact builds and are using other workarounds.

@gnattu , as stated in https://github.com/jellyfin/jellyfin/issues/16549#issuecomment-4186737013, I have used the artifact 7.1.3-4 in my docker as proposed by @verybadsoldier in the exact same way (changing my docker compose and build the image by installing the artifact) and I do reproduce the issue in the exact same way.

I've done cross testing to expose that the artifact 7.1.3-4 has the issue as well :
- Vanilla 10.11.7 ==> Out of Memory
- Vanilla 10.11.7 + DPKG Install jellyfin-ffmpeg7_7.1.3-4-trixie_amd64.deb ==> Out of memory
- Vanilla 10.11.6 ==> OK (took around 200MB of RAM during the process)
- Vanilla 10.11.6 + DPKG Install jellyfin-ffmpeg7_7.1.3-4-trixie_amd64.deb ==> Out of memory

I use docker and give a limit of 5GB to this container.
In all 4 tests I've done, the command from Jellyfin is :
`Starting /usr/lib/jellyfin-ffmpeg/ffprobe with args -analyzeduration 1000000000 -probesize 1000000000 -i file:"/mediacenter/Films/Thor (2011)/Thor (2011).mkv" -threads 0 -v warning -print_format json -show_streams -show_chapters -show_format -show_frames -only_first_vframe`

If I execute manually the exact same command inside the container by removing "threads 0" form arguments, it works without taking too much memory even with the artifact 7.1.3-4.

How I can help you to go further in the debug ?

gnattu:
@ClementToche If you are sure you use the artifact **download from this exact PR**, not the 7.1.3-4 released else where (which does not contain this patch), send your video sample to gnattuoc@me.com

If you did not use the exact artifact, you download it from this page: https://github.com/jellyfin/jellyfin-ffmpeg/actions/runs/23998584430

ClementToche:
@gnattu, following your comment I think you were right ... I was mislead by another comment on the issue.
I did retrieve from scratch the deb package from this specific PR (https://github.com/jellyfin/jellyfin-ffmpeg/actions/runs/23971434238) and redid the test and all is working fine !

Accept my apology for the confusion.

PR #688, opened 2026-04-13
avcodec: add dtsx detection for dts hra

### LINKED ISSUE #16626: DTS-HD HRA encoded audio tracks do not show the DTS:X badge, unlike DTS-HD MA encoded tracks which do
I have a few movies with DTS-HD HRA encoding that carry DTS:X object data, but unlike how Jellyfin displays the "**DTS-HD MA + DTS:X**" label for DTS-HD MA encoded tracks, for these movies it doesn't show any indication as to the DTS:X object data on the audio track label. In fact, it seems like it just entirely omits any indication of the encoding type for DTS-HD HRA tracks outright. (See the attached screenshots for comparison)

A good example of two movies with DTS-HD HRA + DTS:X encoded audio tracks are Fast & Furious 6 and Furious 7 (the 6th and 7th movies in the Fast and Furious collection). While the rest of the series use DTS-HD MA + DTS:X (or Dolby TrueHD + Atmos), these two movies are the only two movies to use HDS-HD HRA audio encoding.

Image
Image
Image

It would be great if Jellyfin could display the DTS:X information in the same was as for DTS-HD HRA as it does for DTS-HD MA. Otherwise, it's not clear from a glance which DTS-HD HRA tracks carry object data and which do not. Granted, not many movies come with DTS-HD HRA encoded audio, but there are a few prominent examples.

gnattu:
The magic sync word is taken from https://github.com/MediaArea/MediaInfoLib/commit/a58e2a2000d0236652dde9e6e0462cb0408aad28

This might be upstream-able. Will send a patch target ffmpeg mainline soon.

**Issues**

Fixes https://github.com/jellyfin/jellyfin/issues/16626

### DIFF
--- a/debian/patches/0088-add-dtsx-detect-for-dts-hd-hra.patch
+++ b/debian/patches/0088-add-dtsx-detect-for-dts-hd-hra.patch
@@ -1,0 +1,114 @@
+Index: FFmpeg/libavcodec/dca_core.c
+===================================================================
+--- FFmpeg.orig/libavcodec/dca_core.c
++++ FFmpeg/libavcodec/dca_core.c
+@@ -2370,8 +2370,14 @@ int ff_dca_core_filter_frame(DCACoreDeco
+         return ret;
+ 
+     // Set profile, bit rate, etc
+-    if (s->ext_audio_mask & DCA_EXSS_MASK)
+-        avctx->profile = AV_PROFILE_DTS_HD_HRA;
++    if (s->ext_audio_mask & DCA_EXSS_MASK) {
++        if (dca->exss.x_imax_syncword_present)
++            avctx->profile = AV_PROFILE_DTS_HD_HRA_X_IMAX;
++        else if (dca->exss.x_syncword_present)
++            avctx->profile = AV_PROFILE_DTS_HD_HRA_X;
++        else
++            avctx->profile = AV_PROFILE_DTS_HD_HRA;
++    }
+     else if (s->ext_audio_mask & (DCA_CSS_XXCH | DCA_CSS_XCH))
+         avctx->profile = AV_PROFILE_DTS_ES;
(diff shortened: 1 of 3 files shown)


Shadowghost (approved):

gnattu:
The windows build failure comes from the recent mingw updates which provides `strndup` support but only in C23 mode, this somehow confuses libass build toolchain which detects `strndup`as available but be unable to use that in actual build. Before the mingw update, `strndup` support is not detected and libass uses its own fallbacks.

Investigating a patch now.

gnattu:
@nyanmisaka When should we add this? If the ffmpeg8 upgrade will be completed very soon I would rather hold this until that is merged first.

nyanmisaka:
> @nyanmisaka When should we add this? If the ffmpeg8 upgrade will be completed very soon I would rather hold this until that is merged first.

The ffmpeg8 upgrade is basically complete, but it still needs some time for testing. We should wait until JF 12.0 RC to release it, otherwise downstream maintainers might adopt it in JF 10.11, potentially leading to regressions.

How about submit it to the upstream for review first? Although I'm not sure who the maintainer is.

gnattu:
Upstream PR is up: https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23013

Shadowghost:
Are we planning to include this in our 8.1 even though we have an upstream PR (which has some comment from their AI bot)

nyanmisaka:
This PR should be harmless since it only adds some new profile strings. Will there be subsequent server PRs that can utilize this?

Shadowghost:
I don't think we need to do anything, since it only has an impact on presentation I think? Codec on ffmpeg side will still be `dca` afaik

nyanmisaka:
Then we just need to rebase this and merge

nyanmisaka (approved):

PR #690, opened 2026-04-17
Remove hardening for arm64

mertalev:
**Changes**
Removing the `--toolchain=hardened` flag makes H.265 decode performance on CPU roughly 30-35% faster. It makes H.264 decode a few percent faster and probably has a modest performance benefit elsewhere as well.

It doesn't seem like amd64 or macOS builds use this flag, so it was specifically penalizing arm64 Linux. I noticed this while running Docker Desktop on a Mac and seeing a huge difference in transcoding performance between the host portable macOS build and the Debian build in the container (4K H.265 -> 4K H.265 transcoding in particular was 5.6x slower).

Some other observations (not directly relevant to the PR, but possibly useful):
* For encoding, setting `-x265-params pools=14` alone had a 4.6x uplift for 4K as it seems the encoder doesn't set up a thread pool automatically on Linux arm64
* The portable Linux arm64 build is ~6% faster than the Debian build as of this PR. Unlike the Debian build, it did not have hardening set and has some performance uplift from better inlining etc.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -61,8 +61,7 @@ CONFIG_x86 := --arch=amd64 \
 	--enable-nvdec \
 	--enable-nvenc \
 
-CONFIG_ARM_COMMON := --toolchain=hardened \
-	--enable-cross-compile \
+CONFIG_ARM_COMMON := --enable-cross-compile \
 	--enable-rkmpp \
 	--enable-rkrga \
 


nyanmisaka (approved):
This option is inherited from [Debian ffmpeg](https://salsa.debian.org/multimedia-team/ffmpeg/-/blob/debian/master/debian/rules?ref_type=heads#L26), which is known for its conservative approach. Since it causes a performance penalty, removing it is reasonable. In any case, we generally avoid using software transcoding on Linux arm64 single-board computers.

The portable version uses the bleeding edge version of toolchains, so it's easy to squeeze out extra performance. Furthermore, libx265 provided by Debian 12 and Ubuntu 24.04 [does not enable assembly on arm64](https://salsa.debian.org/multimedia-team/x265/-/commit/c5a18366e8c5755f3f729776b0de1f07ba26a26b).

Shadowghost (approved):

mertalev:
> The portable version uses the bleeding edge version of toolchains, so it's easy to squeeze out extra performance. Furthermore, libx265 provided by Debian 12 and Ubuntu 24.04 [does not enable assembly on arm64](https://salsa.debian.org/multimedia-team/x265/-/commit/c5a18366e8c5755f3f729776b0de1f07ba26a26b).

I tested on 13 for reference. It's interesting because the portable amd64 build's performance is identical to the deb on my 7800X3D, but transcoding is consistently faster with portable Linux arm64 on my M4 Max.

[Source video is 4K HEVC](https://drive.google.com/file/d/1Ic9DZXMSo07EJMqCFaQRKSSrSw6y1mYv/view?usp=sharing)

| Target Codec | Resolution | deb (main) | deb (PR) | portable (main) | main->PR | PR->portable | main->portable |
|---|---|---|---|---|---|---|---|
| h264 | 720 | 1.42 | 1.12 | 1.07 | -21% | -4.5% | -25% |
| h264 | 1080 | 1.55 | 1.25 | 1.17 | -19% | -6.4% | -25% |
| h264 | 1440 | 1.79 | 1.51 | 1.36 | -16% | -9.9% | -24% |
| h264 | 2160 | 2.10 | 2.00 | 1.74 | -4.8% | -13% | -17% |
| hevc | 720 | 1.54 | 1.22 | 1.17 | -21% | -4.1% | -24% |
| hevc | 1080 | 1.87 | 1.56 | 1.49 | -17% | -4.5% | -20% |
| hevc | 1440 | 2.35 | 2.07 | 1.94 | -12% | -6.3% | -17% |
| hevc | 2160 | 3.57 | 3.37 | 3.08 | -5.6% | -8.6% | -14% |
| av1 | 720 | 1.49 | 1.14 | 1.11 | -23% | -2.6% | -26% |
| av1 | 1080 | 1.80 | 1.48 | 1.41 | -18% | -4.7% | -22% |
| av1 | 1440 | 2.13 | 1.89 | 1.79 | -11% | -5.3% | -16% |
| av1 | 2160 | 2.71 | 2.59 | 2.37 | -4.4% | -8.5% | -13% |

Command (varying codec, bitrate and resolution, removed `scale` for 4K, preset 10 for av1):
  ```bash
  ffmpeg -y -hide_banner -loglevel error \
         -t 2 -i INPUT.mov \
         -vf "scale=-2:1080,tonemapx=tonemap=hable:p=bt709:t=bt709:m=bt709:r=pc:peak=100:format=yuv420p" \
         -c:v libx264 -preset veryfast -b:v 4000k \
         -frames:v 120 -an -g:v 120 -keyint_min:v 120 \
         -f hls -hls_time 2 -hls_list_size 0 \
         -hls_segment_type fmp4 -hls_fmp4_init_filename init.mp4 -hls_flags temp_file \
         -hls_segment_filename seg_%d.m4s out.m3u8
  ```

Bond-009 (approved):

PR #664, opened 2026-03-11
New upstream version 8.1.1

nyanmisaka:
~Extensive testing is required. The Vulkan parts need to be rewritten and are not yet complete.~

**Changes**
- New upstream version 8.1.1
- Update patches for 8.1.1
- Update build scripts and dependencies
- Bump version to 8.1.1-1
- Enable vulkan on Windows clang builds
- Use EOTF from BT 2446 Method B for HLG tonemap

PR #706, opened 2026-05-17
Fix HLG OOTF to use luma-weighted scaling

mertalev:
In [this](https://github.com/immich-app/immich/issues/28408) issue, a user showed that their videos have a reddish hue compared to the original, which happens regardless of tone-mapping algorithm unless they disable tone-mapping completely (no tonemapx). With the sample video they shared, I looked into this with help from Claude. The changes were simplified and optimized by me and tested on an M4 Max, 7800x3D and RTX 4090. The tests used hable and bt2390, but the fixes apply to all algorithms.

I know there's active work on 8.1.1 (which I'm looking forward to!), so I'm happy to base this PR on #664 if you prefer.

## Changes
* tonemap_cuda, tonemap_opencl, tonemap_videotoolbox and tonemapx now all use luma-weighted scaling for OOTF instead of per-channel. This is completely free on all but tonemapx, where it incurs a cost due to the added LUT that ranges from 2.6% if transcoding to medium preset H264 to 12.8% for ultrafast. The cost is only for HLG content since the LUT is not built or used for PQ or SDR.
* tonemapx's `rgb2rgb` now snapshots red, green and blue for computations. Updating `r_lin` in-place and using the new value to update `g_lin` and `b_lin` was wrong. The other filters didn't have this bug.
* A few methods in these filters ended up becoming dead code after the changes, so they were removed.

## Comparisons

<details>
<summary>hable</summary>

### Before
buggy_proper

### After

C_proper

### libplacebo

placebo_hable

</details>

<details>
<summary>BT.2390</summary>

### Before

buggy_bt2390

### After

C_bt2390

### libplacebo

I have no idea why libplacebo BT.2390 looks this bad. BT.2446a and ST 2094-40 look normal and closer to tonemapx BT.2390, so I'm including them below for a broader comparison.

placebo_bt2390

#### BT.2446a

placebo_bt2446a

#### ST 2094-40

placebo_st2094_40

</details>

### DIFF
--- a/debian/patches/0004-add-cuda-tonemap-impl.patch
+++ b/debian/patches/0004-add-cuda-tonemap-impl.patch
@@ -352,7 +352,7 @@ Index: FFmpeg/libavfilter/cuda/colorspace_common.h
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/cuda/colorspace_common.h
-@@ -0,0 +1,341 @@
+@@ -0,0 +1,331 @@
 +/*
 + * This file is part of FFmpeg.
 + *
@@ -409,6 +409,7 @@ Index: FFmpeg/libavfilter/cuda/colorspace_common.h
 +extern __constant__ const float input_uv_scale;
 +extern __constant__ const float output_quantization_factor;
 +extern __constant__ const float output_quantization_scale;
++extern __constant__ const float hlg_gm1;
 +
 +
 +static __inline__ __device__ float get_luma_dst(float3 c, const float3& luma_dst) {
@@ -476,21 +477,6 @@ Index: FFmpeg/libavfilter/cuda/colorspace_common.h
 +    return inverse_eotf_st2084_common(x);
 +}
 +
-+static __inline__ __device__ float ootf_1_2(float x) {
-+    return x > 0.0f ? __powf(x, 1.2f) : x;
-+}
-+
-+static __inline__ __device__ float inverse_ootf_1_2(float x) {
-+    return x > 0.0f ? __powf(x, 1.0f / 1.2f) : x;
-+}
-+
-+static __inline__ __device__ float oetf_arib_b67(float x) {
-+    x = max(x, 0.0f);
-+    return x <= (1.0f / 12.0f)
-+           ? sqrtf(3.0f * x)
-+           : (ARIB_B67_A * __logf(12.0f * x - ARIB_B67_B) + ARIB_B67_C);
-+}
-+
 +static __inline__ __device__ float inverse_oetf_arib_b67(float x) {
 +    x = max(x, 0.0f);
 +    return x <= 0.5f
@@ -498,14 +484,18 @@ Index: FFmpeg/libavfilter/cuda/colorspace_common.h
 +           : (__expf((x - ARIB_B67_C) / ARIB_B67_A) + ARIB_B67_B) * (1.0f / 12.0f);
 +}
 +
-+// linearizer for HLG/ARIB-B67
++// linearizer for HLG/ARIB-B67 (scene-light only; OOTF is applied separately by ootf_hlg)
 +static __inline__ __device__ float eotf_arib_b67(float x) {
-+    return ootf_1_2(inverse_oetf_arib_b67(x)) * (12.0f / REFERENCE_WHITE_ALT);
++    return inverse_oetf_arib_b67(x) * (12.0f / REFERENCE_WHITE_HLG);
 +}
 +
-+// delinearizer for HLG/ARIB-B67
-+static __inline__ __device__ float inverse_eotf_arib_b67(float x) {
-+    return oetf_arib_b67(inverse_ootf_1_2(x / (12.0f / REFERENCE_WHITE_ALT)));
++// BT.2100 luma-weighted OOTF for HLG. Input is display-scale RGB (post 12/REF_HLG factor).
++// hlg_gm1 is baked at PTX link time so __powf can use the constant-exponent fast path.
++static __inline__ __device__ float3 ootf_hlg(float3 c) {
++    float Y_disp = 0.2627f * c.x + 0.6780f * c.y + 0.0593f * c.z;
++    float Y_scene = Y_disp * (REFERENCE_WHITE_HLG / 12.0f);
++    float gain = Y_scene > 0.0f ? __powf(Y_scene, hlg_gm1) : 0.0f;
++    return make_float3(c.x * gain, c.y * gain, c.z * gain);
 +}
 +
 +// delinearizer for BT709, BT2020-10
@@ -698,7 +688,7 @@ Index: FFmpeg/libavfilter/cuda/host_util.c
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/cuda/host_util.c
-@@ -0,0 +1,77 @@
+@@ -0,0 +1,74 @@
 +/*
 + * This file is part of FFmpeg.
 + *
@@ -725,7 +715,7 @@ Index: FFmpeg/libavfilter/cuda/host_util.c
 +#define CHECK_CU(x) FF_CUDA_CHECK_DL(ctx, cu, x)
 +#define DEPTH_BYTES(depth) (((depth) + 7) / 8)
 +
-+int ff_make_cuda_frame(AVFilterContext *ctx, CudaFunctions *cu, int make_cuTex,
++int ff_make_cuda_frame(AVFilterContext *ctx, CudaFunctions *cu, int make_cu_tex,
 +                       FFCUDAFrame *dst, const AVFrame *src, const AVPixFmtDescriptor *src_desc)
 +{
 +    int i, ret = 0;
@@ -738,26 +728,23 @@ Index: FFmpeg/libavfilter/cuda/host_util.c
 +        dst->tex[i] = 0;
 +    }
 +
-+    for (i = 0; make_cuTex && (i < dst->planes); i++) {
-+#ifndef CU_TRSF_NORMALIZED_COORDINATES
-+  #define CU_TRSF_NORMALIZED_COORDINATES 2
-+#endif
++    for (i = 0; make_cu_tex && (i < dst->planes); i++) {
 +        CUDA_TEXTURE_DESC tex_desc = {
 +            .addressMode = { CU_TR_ADDRESS_MODE_CLAMP, CU_TR_ADDRESS_MODE_CLAMP },
-+            .filterMode = i == 0 ? CU_TR_FILTER_MODE_POINT : CU_TR_FILTER_MODE_LINEAR,
-+            .flags = i == 0 ? 0 : CU_TRSF_NORMALIZED_COORDINATES,
++            .filterMode  = i ? CU_TR_FILTER_MODE_LINEAR : CU_TR_FILTER_MODE_POINT,
++            .flags       = i ? 2 /* CU_TRSF_NORMALIZED_COORDINATES */ : 0
 +        };
 +
 +        CUDA_RESOURCE_DESC res_desc = {
-+            .resType = CU_RESOURCE_TYPE_PITCH2D,
-+            .res.pitch2D.format = DEPTH_BYTES(src_desc->comp[i].depth) == 1 ?
-+                                  CU_AD_FORMAT_UNSIGNED_INT8 :
-+                                  CU_AD_FORMAT_UNSIGNED_INT16,
-+            .res.pitch2D.numChannels = i == 0 ? 1 : (dst->planes == 2 ? 2 : 1),
-+            .res.pitch2D.width = i == 0 ? src->width : AV_CEIL_RSHIFT(src->width, src_desc->log2_chroma_w),
-+            .res.pitch2D.height = i == 0 ? src->height : AV_CEIL_RSHIFT(src->height, src_desc->log2_chroma_h),
++            .resType                  = CU_RESOURCE_TYPE_PITCH2D,
++            .res.pitch2D.format       = DEPTH_BYTES(src_desc->comp[i].depth) == 1 ?
++                                        CU_AD_FORMAT_UNSIGNED_INT8 :
++                                        CU_AD_FORMAT_UNSIGNED_INT16,
++            .res.pitch2D.numChannels  = i ? (dst->planes == 2 ? 2 : 1) : 1,
++            .res.pitch2D.width        = AV_CEIL_RSHIFT(src->width,  i ? src_desc->log2_chroma_w : 0),
++            .res.pitch2D.height       = AV_CEIL_RSHIFT(src->height, i ? src_desc->log2_chroma_h : 0),
 +            .res.pitch2D.pitchInBytes = src->linesize[i],
-+            .res.pitch2D.devPtr = (CUdeviceptr)src->data[i],
++            .res.pitch2D.devPtr       = (CUdeviceptr)src->data[i]
 +        };
 +
 +        if ((ret = CHECK_CU(cu->cuTexObjectCreate(&dst->tex[i], &res_desc, &tex_desc, NULL))) < 0)
@@ -807,7 +794,7 @@ Index: FFmpeg/libavfilter/cuda/host_util.h
 +#include "libavfilter/avfilter.h"
 +#include "shared.h"
 +
-+int ff_make_cuda_frame(AVFilterContext *ctx, CudaFunctions *cu, int make_cuTex,
++int ff_make_cuda_frame(AVFilterContext *ctx, CudaFunctions *cu, int make_cu_tex,
 +                       FFCUDAFrame *dst, const AVFrame *src, const AVPixFmtDescriptor *src_desc);
 +
 +#endif /* AVFILTER_CUDA_HOST_UTIL_H */
@@ -1173,7 +1160,7 @@ Index: FFmpeg/libavfilter/cuda/tonemap.cu
 ===================================================================
 --- /dev/null
 +++ FFmpeg/libavfilter/cuda/tonemap.cu
-@@ -0,0 +1,661 @@
+@@ -0,0 +1,669 @@
 +/*
 + * This file is part of FFmpeg.
 + *
@@ -1200,6 +1187,7 @@ Index: FFmpeg/libavfilter/cuda/tonemap.cu
 +extern __constant__ const enum TonemapAlgorithm tonemap_func;
 +extern __constant__ const float tone_param;
 +extern __constant__ const float desat_param;
++extern __constant__ const int lut_size;
 +extern __constant__ const int enable_dither;
 +extern __constant__ const float dither_size;
 +extern __constant__ const float dither_quantization;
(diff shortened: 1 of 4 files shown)


gnattu:
Please ask us first before let AI to invent the thing we already invented.

We DO NOT use per channel scaling despite the LLM told you so. What the user is seeing is the max channel scaling artifact that will make things appears to be too red which is a known limitation and a performance trade off.

This does not happen on any of the hardware accelerated tone mapping path as the default is now ICtCp Intensity based scaling (which is similar to what libplacebo doing, but libplacebo is using a custom IPT-PQ-C4 colorspace doing this) that the color produces is natural enough, definitely not having red tint:

image

We noticed that you are using Reinhard but Jellyfin recommend bt2390 as default.

The above is produced by `tonemap_videotoolbox=format=nv12:p=bt709:t=bt709:m=bt709:tonemap=bt2390:peak=100:desat=0` on 7.1.3.

The red tint problem is currently only happening with the software tonemapx because that is part of the performance trade off.

I'd close this as it is due to misunderstanding on how the filters work.

gnattu:
For reference, we already have a technical comparison on the tone map mode here: https://github.com/jellyfin/jellyfin-ffmpeg/issues/429 and https://github.com/jellyfin/jellyfin-ffmpeg/pull/430

We used to use max channel scaling for all modes (and you can see the red tint in that issue as well) but now we switched to ICtCp based tone mapping on all GPU shaders we write.

mertalev:
I'm sorry if I offended by using AI for this; I just wanted to help land a fix for this issue. Even if the tonemapx LUT is not acceptable, the other filters still benefit from this change and tonemapx has a separate bug that this PR also fixes. 

Also, the colors are definitely different with this PR with tonemap_videotoolbox and (in my opinion) closer to the original.

Before (hable):

vtb_baseline_hable

After (hable):

vtb_v4_hable

Before (bt2390):

vtb_baseline_bt2390

After (bt2390):

vtb_v4_bt2390

> Also even the software part your LLM does not change the place where it really matters(the sims intrinsics part), so I'd close this as this does not really work and it is due to misunderstanding on how the filters work.

I tested both NEON and AVX2 for reference (in addition to all the GPU filters). All of the backends were updated in this PR.

gnattu:
> I just wanted to help land a fix for this issue. Even if the tonemapx LUT is not acceptable, the other filters still benefit from this change and tonemapx also has a separate bug that this PR also fixes.

It is always hard to accept PRs from author lacks necessary understanding and LLMs usually make things worse because it would even attempt to "fix" something with code that is totally irrelevant and claim it was the fix.

For tonemapx, the "separate bug" you mentioned actually helps more to the red tint than the ootf you think was the major problem.

So what I do prefer is, discard the ootf changes for all filters for now because that is more of a behavior change rather than a bug fix, and only contain the tonemapx matrix multiplication fix so that we can accept it as-is. This should reduce the red tint on software filter on warm scenes a lot already.

mertalev:
> For tonemapx, the "separate bug" you mentioned actually helps more to the red tint than the ootf you think was the major problem.

That was the first change and it did help, but the OOTF change was because it was still more red than the original and libplacebo. The OOTF change actually made a bigger difference in R/G ratio (IIRC it started at 1.27 -> 1.24 with the bug fix -> 1.17 with luma-scaling).

I do again apologize and totally understand being reluctant about LLM changes (I am too as a maintainer). I did put effort into this with statistical validation, perf flame graphs, kernel timings, etc., but ultimately I don't have the background or context that you do. I personally think hable on VT looks okay without the change while bt2390 looks a bit oversaturated.

I should mention that I really did mean that the changes are free for the other filters - they are microseconds faster if anything (the difference is invisible while transcoding though). If you still prefer, I can de-scope this PR to only the tonemapx bug fix without any of the OOTF changes. Just let me know how you'd like to proceed and I'll remove the other changes.

gnattu:
> but the OOTF change was because it was still more red than the original and libplacebo

It is expected but it does have performance impact like I mentioned, but the impact should be smaller on GPU which makes it a more feasible change on GPU shaders, but I really don't think it could land in tonemapx now as the performance penalty for current implementation is too a bit too high.

I *think* you can keep OOTF enhancements for GPU filters, but only apply matrix fix for tonemapx. We already made some quality trade offs for that filter and this can just be another one.

For the GPU shader we prefer not using if branches as that could be expensive on GPUs, with the exception on Metal with comparison solely to function constants, where the complier would remove the dead branch entirely to also remove the performance impact. I noticed that in current implementation there are a few if branches added and I don't think is necessary at all.

mertalev:
Sounds good! I'll remove the tonemapx OOTF changes and rework the GPU filters to not add runtime branches.

Possibly useful note: while working on tonemapx, I tried a few different approaches including no LUT, a smaller linear-interpolated LUT and a smaller non-interpolated LUT. The first two were slower and the only thing faster was using smaller LUTs with bit-shifting since it relieved L1 cache pressure. At the extreme end, a 4K `lin_lut`, 4K `tonemap_lut` and 1K `ootf_lut` reduced the overhead added in this PR by 70%, at the cost of some quantization errors. It actually looked fine visibly, but zooming into deep shadows showed some posterization.

nyanmisaka:
I think adding a separate ootf pass complicates things. The current hlg eotf is derived from zimg a few years ago, and it seems to cause some saturation issues from your images because it's not scene-referred. A better approach would be to switch to the eotf used by libplacebo, which is scene-referred and combines inverse-oetf and ooft in a single pass, resulting in minimal overall changes. I can add this change to the 8.1 branch later.

mertalev:
That sounds promising! With that, I assume none of the OOTF-related changes in this PR are needed. Should I still change this branch to only include the tonemapx bug fix, or would you like to take it over from here for the 8.1 branch?

nyanmisaka:
I've pushed out an update that includes everything that's needed, you can try it and see if it helps.

mertalev:
That was quick! I tested tonemapx and VT with the CI binary.

tonemapx seems to have a bug since it gets a yellow cast.

I think the VT colors look good, but the brightness seems elevated. The new brightness looks great with hable (which was dark before), but bt2390 looks bright now (although libplacebo's bt2390 is also bright compared to bt2446a/st2094-40).

For reference, here are the Y values:

| Algorithm | ff81 tonemapx | ff81 VT | libplacebo |
|---|---:|---:|---:|
| linear | 106.22 | 107.72 | 113.88 |
| hable | 129.30 | 134.06 | 126.61 |
| gamma | 138.04 | 145.08 | 144.30 |
| reinhard | 149.87 | 157.05 | 150.59 |
| bt2390 | 152.71 | 159.40 | 152.08 |
| mobius | 152.63 | 159.70 | 152.35 |
| bt2446a | - | - | 130.86 |
| st2094-40 | - | - | 128.21 |
| spline | - | - | 124.48 |


<details>
<summary>tonemapx</summary>

#### hable

ff81_tonemapx_hable

#### BT.2390
ff81_tonemapx_bt2390

</details>

<details>
<summary>tonemap_videotoolbox</summary>

#### hable

ff81_vt_hable

#### BT.2390

ff81_vt_bt2390

</details>

gnattu:
I could look at the brightness later because I do believe the normalized brightness of libplacebo is a bit off and our old one is more sane.

mertalev:
FWIW I tried adding bt2446a to VT in the 8.1 branch. The Y is 132.48 and the result visually looks the best out of everything from tonemapx or VT so far (libplacebo bt2446a is my top pick overall though).

ff81_vt_bt2446a

gnattu:
bt2446a requires very accurate metadata (and mastering) and it will be too dark for a lot of use cases, which makes it less attractive to be added to Jellyfin, at least not as a default. If you search upstream libplacebo or mpv you could find a lot of bad examples it demonstrates.

nyanmisaka:
> tonemapx seems to have a bug since it gets a yellow cast.

This should be fixed in the latest push.

mertalev:
> bt2446a requires very accurate metadata (and mastering) and it will be too dark for a lot of use cases

Oof, you're right - it drops like a stone to 113.97 at `peak=200`. I think it's still good as an option, but risky as a default as you say.

> This should be fixed in the latest push.

Yes, the tonemapx colors look better now. It's more red than VT (to be expected), but not overtly red like in the issue. Brightness is a bit less than VT, which helps for bt2390 (both tonemapx and VT are a bit too bright here though).

<details>
<summary>tonemapx</summary>

#### hable

fixed_tx_hable

#### BT.2390

fixed_tx_bt2390

</details>

#### Edit

I did some more testing with libplacebo (since it seems to already have every option under the sun). My takeaway is that bt2390 just doesn't work well for this content, no matter the configuration, and the same with mobius and reinhard. hable is okay, but I think that's more because it's just darker in general and that happens to balance things out here.

I actually think libplacebo's bt2446a and st2094-40 work pretty well on all the content I tested so far, possibly because its peak detection makes them more resilient to inaccurate metadata. I found that `peak_percentile` is a very strong lever to adjust brightness as well, so changing from the default 99.995% to e.g. 99.9% is enough to brighten it up depending on preference (lowering the percentile thankfully doesn't seem to change too much for the user's video). I did a PoC for percentile-based peak detection with VT and saw no FPS impact, so I do think there's potential here. If you're open to it, I can make an enhancement issue for this.

#### Edit 2

I apparently mixed up the images for BT.2446 and ST 2094-40 (including in the OP) - I actually slightly prefer ST 2094-40. I also didn't look at spline before, which I would say is similar to ST 2094-40 but a touch darker.

gnattu:
I actually think we should bypass the old PQ oriented tone mapping and just use HLG native 1000nit to 203nit OOTF for brightness compression as that probably produces the most natrual results:

image

And this approach is even faster, as it is now a simple OOTF pass instead of doing the whole tone map filter.

> My takeaway is that bt2390 just doesn't work well for this content, no matter the configuration, and the same with mobius and reinhard.

And yes, I do agree here because the HLG EOTF converted to liner-space tends to have too much area of signals close or higher than diffuse white for a lot of input sources which will cause those tone mappers to over-compress and generate subpar results. HLG signals are scene based not display based after all.

I checked the new libplacebo migrated EOTF and it is mathematically correct and actually more accurate than our old one. It is just the tone mappers works bad with HLG signals this way.

> I actually think libplacebo's bt2446a and st2094-40 work pretty well on all the content I tested so far, possibly because its peak detection makes them more resilient to inaccurate metadata.

This is not the case though, peak detection is actually gambling in real work use cases as some source works great with that on but some will cause weird tone mapped brightness as well. And HLG is scene based which means it does need "metadata" for display. The two works well more likely because they are naturally darker which workarounds the over bright nature of HLG sources.

mertalev:
> I actually think we should bypass the old PQ oriented tone mapping and just use HLG native 1000nit to 203nit OOTF for brightness compression as that probably produces the most natrual results:

That seems interesting. The main downside seems to be that the tulips have flatter colors with less highlights, but the overall image looks natural. [This](https://downloads.bbc.co.uk/rd/pubs/papers/HDR/BBC_HDRTV_HLG_LUT_Production_Guide.pdf) BBC paper has some recommendations for HLG based on content if it helps.

> This is not the case though, peak detection is actually gambling in real work use cases as some source works great with that on but some will cause weird tone mapped brightness as well. And HLG is scene based which means it does need "metadata" for display. The two works well more likely because they are naturally darker which workarounds the over bright nature of HLG sources.

Yeah, I can see how it can be unpredictable. There was one dim scene I saw where setting 99.9% or 99.95% looked fine, but then using 99.99% *completely* changed the brightness even though the rest of the video was pretty similar for all of them.

nyanmisaka:
The brightness issue should have been improved in the latest push. The mistake was that I placed get_luma_src() before HLG's inverse_oetf.

gnattu:
Oh, I only checked if the EOTF math and didn't check the luma source.

But that does not make BT2390 look much better:

image

As the signal itself has too many area around and above diffusion white brightness, the tone mapped is still subjectively too bright.

Native HLG OOTF still produces "more natural" results IMO though:

image

It is not as bright but perceptually closer to what users are expecting from an SDR source.

nyanmisaka:
> Oh, I only checked if the EOTF math and didn't check the luma source.
> 
> But that does not make BT2390 look much better:

> As the signal itself has too many area around and above diffusion white brightness, the tone mapped is still subjectively too bright.
> 
> Native HLG OOTF still produces "more natural" results IMO though:

> It is not as bright but perceptually closer to what users are expecting from an SDR source.

Yes, even using bt2390 and peak_detect=1 in libplacebo didn't yield the desired brightness in this frame.

I don't have enough HLG samples to verify that skipping the tonemapper is reliable, but this should be a last resort without introduing bt2446a or st2094-40. What changes are needed to achieve this result?

mertalev:
The user shared another video [here](https://github.com/immich-app/immich/issues/28408#issuecomment-4478543170) of a choir, which can give a reference for skin tones. The videos are shot from a Pixel phone, so I imagine any HDR content from a Pixel (or iPhone if omitting DoVi) would be usable for testing.

suuuehgi:
Wow, I appreciate all of your work! ❤️

I am the user who triggered all of this. If I can contribute in any way, please let me know!

I'm not going to interfere in the technical discussion, as I don't know enough about Jellyfin-FFmpeg and color spaces.

The flower image I shared might not have been the best choice, as it's very dynamic and at least I am not the best judge of flower colors.
But as a human I'm very good at reading human faces. And the resulting skin tones of the choir recording (initial pictures [in the other thread/issue](https://github.com/immich-app/immich/issues/28408#issue-4437842679)) -- which, unfortunately, you can only guess at because of the blurring -- were somewhere between "salmon pink" and "just having run a 1,000-meter sprint in the peak of summer" while the wood paneling had changed from brown to bright red.

All the recordings were simple video recordings using my Pixel phone (Pixel 8 Pro to be precise, but I don't think it matters) with (10-bit) HDR switched on -- the standard stock photo/video app.

Unfortunately, hardware acceleration is not accessible on the VPS I run this on.

gnattu:
> Unfortunately, hardware acceleration is not accessible on the VPS I run this on.

Luckily that is what my trick, direct HLG OOTF, can help.

Even with the OOTF performance trade off(we use per channel LUT instead of computing luma), a direct 1000nit to 203nit OOTF pass provides good enough results on tonemapx.

Both are using already fixed matrix with modified tonemapx:

### BT.2390

https://github.com/user-attachments/assets/9a249b4f-cde4-4eb8-be62-9300f92713f4

### Direct HLG OOTF

https://github.com/user-attachments/assets/52d6d85f-5e42-4fdc-b2d4-3a49f78dd072

And because using direct HLG OOTF is reducing computing, not adding, the performance for HLG videos are actually better if we use this approach.

mertalev:
I think the choir video looks good and similar to the bt2446a output in this case:

https://github.com/user-attachments/assets/0eba5d0c-b142-4db3-826c-1212992334fd

I found [this](https://www.youtube.com/watch?v=vv4adaah_Z0) video (downloaded as 4K HDR), which is a good test as it contains many scenes in different lighting and isn't particularly well-mastered. I'm not sure how the OOTF-only videos were made, so I can't speak for how it performs here, but I ran the video through some of the algorithms, posting the frame at 04:16 below. These are unfortunately screenshots because PNG frames from FFmpeg seem to get interpreted as sRGB when I upload them here.

I can say that bt2390 is too bright here as well and ends up clipping a lot of detail in some frames, along with strange coloring at times for highlights like skies and sunlit bricks:

bt2390

bt2446a has very good colors and details, but it has a moodier take on the video:

bt2446a

st2094-40 is in between the two with good colors and contrast, but it loses some highlight detail:

st2094-40

spline is somewhere in between bt2446a and st2094-40. It's tonally similar to st2094-40, but prioritizes highlight detail with slightly lower brightness:

spline

Overall, all of these besides bt2390 give reasonable results for this video depending on preference, with spline seeming like the most well-rounded. I'm curious to see how the OOTF-only approach stacks up here. Also, all of these were with `libplacebo=tonemapping=<algo>:colorspace=bt709:color_primaries=bt709:color_trc=bt709:range=tv:format=yuv420p`, so mostly default libplacebo settings.

gnattu:
With more research I think I find a better approach to this problem as I "fixed" BT.2390.

See https://github.com/nyanmisaka/jellyfin-ffmpeg/pull/15 for more info.

Both libplacebo and us was using the HLG standard convention Lw=1000nit and gamma of 1.2 as EOTF parameters to linearize the HLG input, but the brightness compression of HLG is very different from PQ which caused such linearization became overly bright after using common tone mappers. 

By adopting recommendations from ITU report BT.2446-1, section 5.2.1:

> Rather than using the usual bridge condition of 1 000 cd/m2 for format conversions, the HLG display
EOTF is conveniently set to a nominal peak luminance, Lw, of 291 cd/m2, as that renders “HDR
Reference White” at 78 cd/m2, which is equivalent to 90% of the SDR signal on a BT.1886 reference
display with nominal peak luminance of 100 cd/m2. By doing so, the brightness of the tone mapped
images are comparable to SDR graded content, and a direct 1:1 mapping between HDR displayed
light and SDR displayed light can be used for the conversion, over most of the signal range. 

> The exact HLG gamma display gamma may be adjusted slightly from the value specified for HDR
images in Recommendation ITU-R BT.2100, to achieve the desired artistic effect. But a gamma value
of 1.03 for an EOTF Lw of 291 cd/m2 has been found to work well and complements the inverse tone
mapper described in § 5.1.5.

By doing this we can apply the common tone map operators on it and still get natural results.

This is described in BT2446 Method B, so very different from bt2446a implemented by libplacebo, which still uses Lw=1000 and 1.2 gamma.

This approach also requires minimal code change and performance impact, which means it can be easily applied to tonemapx and produce acceptable results even with the peformance tradeoffs in place:

image

mertalev:
@gnattu That's a great find! The Italy image looks much better than before. If I have to nit, the colors seem a bit muted compared to the ones I posted. It's most visible in the plants as they have both sunlit and shaded leaves intermixed. Would it make sense to tweak the parameters?

gnattu:
> colors seem a bit muted compared to the ones I posted

That part is more of personal preferences, as compressing dynamic range has to compresses color in some way, and there isn't a universally best method for doing so. Even in the BT2446 report, there are three different ways and each produces vastly different results.

HLG EOTF from BT2446 Method B is picked because it provides a reasonable compatibility for tone map operators currently implemented. Not only bt2390, other methods like hable and reinhard also works with this EOTF, and that makes HLG sources less different than others where users can still control the behavior with parameters already there. Another important reason is that this approach has no performance impact which is important for jellyfin's real time transcoding use case.

"Only tone map operator X works for this kind of source" or "skip tone map and force it use internal SDR compatibility (the direct OOTF approach)" are both not ideal for jellyfin use cases. Dynamic tone map operators like st2094-40 and spline are both more complicated and performance intensive, but current static tone map operators already made some lower end GPUs to struggle, even if we eventually implemented those in our ffmpeg, we still need to provide a way that slower hardwares can use and still producing acceptable results.

mertalev:
That's definitely fair! I was just curious if e.g. increasing the gamma would liven up the image. bt2390 has no issues producing rich colors, so it isn't an issue of tone map operator at this point. On that note, st2094-40 in particular is a pain to implement and probably more intensive, while bt2446a and spline are comparatively simple algorithms. I do agree though that it isn't ideal for only a certain operator to work for certain sources.

Issue #718, opened 2026-06-05
[FFmepg 8.1.1-2] Attachements with Jellyfin server 10.11.10 still break


BlackstormFirst:
- **Platform:** [Windows]
- **Operating System:** [Windows 11]
- **Virtualization**: [Vmware Esxi 8]
- **Jellyfin Version:** [e.g. 10.11.10]
- **Jellyfin-FFmpeg Version:** [8.1.1-2]
- **Hardware Acceleration:** [Nvidia]
- **CPU & GPU Model**: [i7-7700K]

Hello with the current pre-release of Ffmpeg 8.1.1-2 the attachements are not extracted.

<img width="816" height="85" alt="Image" src="https://github.com/user-attachments/assets/5f6cf6f7-eeb9-4d4d-af6a-eae53f3eb5f9" />

Close the bug feeback if it's a known issue with this release and have no impact with the next Jellyfin 10.12.0 releases.

gnattu:
This is not helpful. Can you at least provide some logs and if you can find the file that failed the extraction. Reporting like this does not help anyone to fix your problem

BlackstormFirst:
Hi, i hope this will be helpful:

```
[2026-06-06 13:12:56.586 +02:00] [INF] [40] MediaBrowser.MediaEncoding.Attachments.AttachmentExtractor: "ffmpeg" "-dump_attachment:t \"\" -y  -i file:\"Z:\Videos\Animes\Animes VF\Tales of Zestiria the X\Tales of Zestiria the X - 00x01 - Dawn of the Shepherd - 2015 - 1080p - x265 - BDRip - VOSTFR.mkv\" -t 0 -f null null"
[2026-06-06 13:12:56.738 +02:00] [INF] [40] MediaBrowser.MediaEncoding.Attachments.AttachmentExtractor: ffmpeg attachment extraction completed for "file:\"Z:\Videos\Animes\Animes VF\Tales of Zestiria the X\Tales of Zestiria the X - 00x01 - Dawn of the Shepherd - 2015 - 1080p - x265 - BDRip - VOSTFR.mkv\"" to "E:\ProgramData\Jellyfin\Server\data\attachments\a1\a1959c32-d5a0-9e18-ee65-92fd9f74691c"
[...]

PR #719, opened 2026-06-06
[8.1] Relax to allow safe filenames in mkv attachments

nyanmisaka:
This isn't the first time the upstream has introduced more bugs in an attempt to fix a hypothetical vulnerability.

**Changes**
- Relax to allow safe filenames in mkv attachments

**Issues**
- https://github.com/jellyfin/jellyfin-ffmpeg/issues/718


gnattu (approved):
```
+        /* Block control characters and dangerous path characters */
```

Trust me I have seen those in mkv attachments as well

mkv and its tools have little to no restrictions on the filename and a lot of people choose to abuse such flexibility.

nyanmisaka:
> ```
> +        /* Block control characters and dangerous path characters */
> ```
> 
> Trust me I have seen those in mkv attachments as well
> 
> mkv and its tools have little to no restrictions on the filename and a lot of people choose to abuse such flexibility.

`(unsigned char)*f < 32` or `*f == '\\' || *f == ':'`
Are you referring to the former or the latter? The latter is allowed on Unix but not on Windows and is also insecure.

In this patch, we will skip attachment filenames that fail the check instead of letting ffmpeg fail and exit.

Shadowghost (approved):

gnattu:
> Are you referring to the former or the latter? The latter is allowed on Unix but not on Windows and is also insecure.

The latter. 

And for platform compatibility. I have seen something worse. Some files included a very long filename that is supported on some filesystem but exceeds max file length limit on something else. mkv attachment filename was causing and will keep cause troubles, so be prepared for even more weird files incoming.

nyanmisaka:
I can see the potential. If anyone complains, we'll revisit.

PR #723, opened 2026-06-11
feat: add riscv64 (RISC-V) Debian/Ubuntu package and portable build s…

rockyzhang:
…upport

Portable build (cross-compiled via crosstool-ng):
- Add builder/images/base-linuxriscv64/ with Dockerfile, ct-ng-config, gen-implib.sh, cross.meson, toolchain.cmake
- Add builder/variants/linuxriscv64-gpl.sh variant file
- Add build-linux-riscv64 build script
- Add cross-riscv64.meson and toolchain-riscv64.cmake top-level files
- Clear STAGE_CFLAGS (-fvisibility=hidden is shared-lib only and breaks configure scripts for static builds)
- Keep sysroot static libs so cross-compiler can link test programs
- Add -fvisibility=hidden to gen-implib.sh directly (prevents trampoline symbol leakage between X11 import libraries)

crosstool-ng config for RISC-V GCC 15.2.0:
- glibc 2.42, binutils 2.46.0, Linux headers 6.1.100

CI / debian build:
- Add riscv64 to build_portable_linux, jammy/noble/resolute CI jobs
- Add jammy-riscv64 to build.yaml packages list
- Add riscv64 cross-compilation to build, docker-build.sh, debian/rules
- Fix libatomic.so.1: install libatomic1:riscv64, add -latomic LDFLAGS (RISC-V needs libatomic for 128-bit atomics)

Architecture-specific build fixes:
- FFTW3: add linuxriscv64 arch case (no SSE/AVX SIMD)
- libglvnd: disable asm for riscv64 (no RISC-V stubs)
- libva: add linuxriscv64 target (DRM+X11, riscv64 driver path)

FFmpeg configure:
- Disable GPU-hardware features: ffnvcodec/cuda/nvenc (NVIDIA), amf (AMD), libvpl (Intel) — no RISC-V GPU hardware
- Keep: libplacebo, vulkan, shaderc, opencl (libs exist at build time)

**Changes**

**Issues**

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -40,15 +40,7 @@ jobs:
     uses: ./.github/workflows/_meta.yaml
     with:
       distro: 'ubuntu'
-      codenames: '["jammy"]'
-      architectures: '["amd64", "arm64", "riscv64"]'
-      release: false
-
-  build_ubuntu_riscv64:
-    uses: ./.github/workflows/_meta.yaml
-    with:
-      distro: 'ubuntu'
-      codenames: '["noble", "resolute"]'
+      codenames: '["jammy", "noble", "resolute"]'
       architectures: '["amd64", "arm64", "riscv64"]'
       release: false
 
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -17,21 +17,39 @@ on:
     paths-ignore:
       - '**/*.md'
 
+  workflow_dispatch:
+
 jobs:
   build_debian:
     uses: ./.github/workflows/_meta.yaml
     with:
       distro: 'debian'
-      codenames: '["bullseye", "bookworm", "trixie"]'
+      codenames: '["bullseye", "bookworm"]'
       architectures: '["amd64", "arm64"]'
+      release: false
+
+  build_debian_trixie:
+    uses: ./.github/workflows/_meta.yaml
+    with:
+      distro: 'debian'
+      codenames: '["trixie"]'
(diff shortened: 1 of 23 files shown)


.github/workflows/build.yaml
```diff
@@ -17,21 +17,39 @@ on:
...
+      codenames: '["jammy"]'
+      architectures: '["amd64", "arm64", "riscv64"]'
+      release: false
+
+  build_ubuntu_riscv64:
+    uses: ./.github/workflows/_meta.yaml
+    with:
+      distro: 'ubuntu'
+      codenames: '["noble", "resolute"]'
+      architectures: '["amd64", "arm64", "riscv64"]'
```
Shadowghost, line 44: Why two workflows if both build the same architectures?

gnattu:
> Disable GPU-hardware features: ffnvcodec/cuda/nvenc (NVIDIA), amf (AMD), libvpl (Intel) — no RISC-V GPU hardware
Keep: libplacebo, vulkan, shaderc, opencl (libs exist at build time)

Excuse me but why?

I am not sure if libplacebo and its other GPU compute friends are useful on this target.

nyanmisaka:
I'm just unsure if it's worthwhile to introduce FFmpeg builds before .NET fully supports RISCv64. This means we'll have to maintain platforms used by very few people (mostly developers), and we might not even have enough hardware ourselves to validate the binaries.

Wouldn't it be a better option for the maintainers of RISCv64 based Linux distros to apply our patches to their builds for these niche use cases? Of course, this situation might change in the future.

PR #726, opened 2026-06-18
avfilter/d3d11: add tonemap/transpose/deint/scale filters for Qualcomm

gnattu:
**Changes**

This adds the required d3d11 filters for jellyfin use cases. Although d3d11 is a generic API on windows, the filters are implemented in a way that mainly targets Qualcomm SoC use cases and is not intended for general use. 

Qualcomm has a lot of platform specific limitations that affects the implementation of the filters:

- d3d11 video processor does have native deinterlance API, but Qualcomm did not implement that. The deinterlance filter is ported from yadif/bwdif CUDA filters instead
- Qualcomm implemented d3d11 video processor native scaling, but that scaler can only output nv12 frames. To output p010 frame for tone mapping, a shader based bicubic scaler ported from CUDA is used as a fallback
- Qualcomm also implemented d3d11 video processor native overlay, but it is useless for jellyfin use case because it does not handle alpha at all. A shader based overlay filter is implemented instead.
- The tonemap filter is also shader based to handle Dolby Vision.
- The zero copy behavior of D3D11 pipeline is very unstable. A true zero copy is impossible with current driver as the encoder would end up outputting broken frames if a compute shader directly manipulate the texture the encoder is going to use. We have to a least allocate one private texture in between the filter chain and do a GPU side blit to prevent the encoder outputting the broken frame. Current plan is to do this extra copy on the last compute filter before the encoder.

D3D11 filters still performs better than OpenCL, despite all the quirks mentioned above. Given that even the d3d11 zero copy is not reliable, I personally have very low expectation that Qualcomm would fix OpenCL or Vulkan at any time soon.

**Issues**


nyanmisaka:
> The zero copy behavior of D3D11 pipeline is very unstable. A true zero copy is impossible with current driver as the encoder would end up outputting broken frames if a compute shader directly manipulate the texture the encoder is going to use. We have to a least allocate one private texture in between the filter chain and do a GPU side blit to prevent the encoder outputting the broken frame. Current plan is to do this extra copy on the last compute filter before the encoder.

- Is frame synchronization correct between D3D11 filters?
- Is frame downloaded from D3D11 filters also affected?

If the above problems do not exist, it means the MF encoder has created its own internal D3D11 context, thus not sharing the decoder/filter's D3D11 context and the implicit synchronization mechanism is lost. In this case, it may be necessary to manually introduce `flush` and `getdata` to make the CPU wait for the GPU to finish writing to the texture. This allows the MF encoder to retrieve synchronized frames.

nyanmisaka:
Are there any performance metrics to refer to? For example, when using the fastest preset of the MF encoder, can the transcoding speed approach the hardware limit?

gnattu:
> Is frame synchronization correct between D3D11 filters?

It is quite strange because there is no frame corruption if only d3d11 video processor is used, and only compute shader will cause this.

> Is frame downloaded from D3D11 filters also affected?

No it does not. I have tested.

nyanmisaka:
> > Is frame synchronization correct between D3D11 filters?
> 
> It is quite strange because there is no frame corruption if we only d3d11 video processor is used, and only compute shader will cause this.

Does this also apply to GPUs on the x86 platform?

gnattu:
> > > Is frame synchronization correct between D3D11 filters?
> 
> > 
> 
> > It is quite strange because there is no frame corruption if we only d3d11 video processor is used, and only compute shader will cause this.
> 
> 
> 
> Does this also apply to GPUs on the x86 platform?

Have not tested this on x86 yet.

gnattu:
> Are there any performance metrics to refer to? For example, when using the fastest preset of the MF encoder, can the transcoding speed approach the hardware limit?

Current implementation is mainly tested on X1 Plus device which has a super slow encoder. Will test on faster X1 Elite and X2 Elite devices later this week.

nyanmisaka:
> > Are there any performance metrics to refer to? For example, when using the fastest preset of the MF encoder, can the transcoding speed approach the hardware limit?
> 
> Current implementation is mainly tested on X1 Plus device which has a super slow encoder. Will test on faster X1 Elite and X2 Elite devices later this week.

The VPU performance on the X2 is significantly improved, and it may benefit even more from the native d3d11 filter. I wonder if they have addressed the issue of the MF HEVC encoder not supporting Main10 profile. This could be a problem when we enable to 10-bit/HDR encoding in the future.

nyanmisaka:
I tested setting `force_output_copy=0` for both the overlay and tonemap on NVIDIA and didn't notice any frame corruption. It might just be another quirk of the Qcom driver...

```
[h264_mf @ 000001f5559228c0] MFT name: 'NVIDIA H.264 Encoder MFT'
[hevc_mf @ 0000023e68b87980] MFT name: 'NVIDIA HEVC Encoder MFT'
[av1_mf @ 0000016ad0d72e40] MFT name: 'NVIDIA AV1 Encoder MFT'
...

[Parsed_scale_d3d11_2 @ 0000021461009080] D3D11 scale config: 1920x1080 -> 1920x1080
[hwupload @ 000002145e60a180] Surface format is bgra.
[Parsed_overlay_d3d11_3 @ 0000021461008d40] D3D11 shader output: direct UAV
[Parsed_overlay_d3d11_3 @ 0000021461008d40] [framesync @ 00000214597f7f08] Sync level 2
[Parsed_scale_d3d11_2 @ 0000021461009080] Configuring D3D11 video processor: 1920x1080 -> 1920x1080
[Parsed_scale_d3d11_2 @ 0000021461009080] D3D11 video processor successfully configured
[Parsed_overlay_d3d11_3 @ 0000021461008d40] D3D11 shader main input: direct SRV
[Parsed_overlay_d3d11_3 @ 0000021461008d40] D3D11 shader overlay input: direct SRV
...

[Parsed_tonemap_d3d11_0 @ 000002327663e180] D3D11 shader output: direct UAV
[Parsed_tonemap_d3d11_0 @ 000002327663e180] Computed DOVI signal peak: 4.376948 at pts 0
[Parsed_tonemap_d3d11_0 @ 000002327663e180] D3D11 shader input: direct SRV
[Parsed_tonemap_d3d11_0 @ 000002327663e180] Compiling D3D11 tonemap shader variant: tone=7 mode=3 in_trc=16 out_trc=1 dovi=1
```

nyanmisaka:
I've also encountered frame corruption in a hybrid zero-copy pipeline of D3D11VA decoder + QSV filter + MF encoder. The first two are known to work reliably. Therefore, the problem lies with the MF encoder.

Checking `mfenc.c`, it seems the issue stems from the MF encoder's [async encoding](https://github.com/FFmpeg/FFmpeg/blob/bccaf86a89699ffb54b01503c935eb489a39a8b0/libavcodec/mfenc.c#L1175). For software frames, a [deep copy](https://github.com/FFmpeg/FFmpeg/blob/bccaf86a89699ffb54b01503c935eb489a39a8b0/libavcodec/mfenc.c#L444) of each frame is performed before encoding, while for DX11 hardware frames, only the texture of each frame is [referenced/wrapped](https://github.com/FFmpeg/FFmpeg/blob/bccaf86a89699ffb54b01503c935eb489a39a8b0/libavcodec/mfenc.c#L392) before encoding.

This means that before the encoder consumes the DX11 frame, it might have already been released to the DX11 frame pool. This causes the DX11 filter to write frames that the MF encoder is currently referencing, resulting in frame corruption or out-of-order display.

gnattu:
What is the corruption you are seeing? For qualcomm, the corruption is either the output frame turns all green, or the video is flipping between green frames and seems to be okay frames.

nyanmisaka:
> What is the corruption you are seeing? For qualcomm, the corruption is either the output frame turns all green, or the video is flipping between green frames and seems to be okay frames.

Most of the problematic frames are out-of-order, with very few being frame tearing.

gnattu:
> Most of the problematic frames are out-of-order, with very few being frame tearing.

It is not the same failure mode then.

nyanmisaka:
Well, it seems I've discovered a different failure mode in the Intel driver's MF impl, as there are no failures at all in the NVIDIA driver. However, the async encoding in `mfenc.c` doesn't seem robust enough in any case.

For multi-GPU environments, this patch is also needed. Otherwise, the MF runtime may cause ffmpeg to crash. 
[0001-avcodec-mfenc-filter-hw-encoder-by-d3d11-adapter-LUI.patch](https://github.com/user-attachments/files/29161816/0001-avcodec-mfenc-filter-hw-encoder-by-d3d11-adapter-LUI.patch)

gnattu:
I have ran speedtest on faster devices, using our official 4K test videos at https://repo.jellyfin.org/test-videos/

All outputs are 2160p mp4  with `h264_mf -hw_encoding 1 -compression_level 0` and 30M bitrate.

The performance characteristics is quite strange on Qualcomm VPU as the HEVC decoding performance does not look consistent to me, it looks like, depending on the source file, the performance can have huge difference, especially on X2 Elite. 

I have no idea why HDR10 and Dolby Vision HEVC sources has this huge fps difference.

The scale filter only does `format=nv12` so it is not even using any compute shaders. At least I *think* our compute shader's GPU side blit would not bottleneck the VPU.

The overlay uses `sdr_hevc8` as input but overlay cannot run at full encoder speed on a lot of platforms due to its CPU and GPU hybrid nature.

| Test | X2 Elite fps | X1 Elite fps | Speedup |
|------|------------:|-------------:|------:|
| sdr_avc_direct       | 193 |  88 | 2.19x |
| sdr_hevc8_direct    | 223 | 100 | 2.23x |
| sdr_hevc10_scale    | 135 |  86 | 1.57x |
| sdr_av1_scale       | 163 |  83 | 1.96x |
| sub_ass_overlay     | 134 |  91 | 1.47x |
| sub_srt_overlay     | 144 |  91 | 1.58x |
| hdr10_hevc_tonemap  | 132 |  64 | 2.06x |
| hdr10_av1_tonemap   | 185 |  81 | 2.28x |
| dv_p5_tonemap       | 194 |  84 | 2.31x |

gnattu:
> However, the async encoding in mfenc.c doesn't seem robust enough in any case

Unfortunately this is what all Qualcomm can use. It does not have vendor specific API like other chips on Windows, it also has not implemented any of the d3d12va encoders.

> For multi-GPU environments, this patch is also needed. Otherwise, the MF runtime may cause ffmpeg to crash.

For Qualcomm devices this is not a typical use case though, there is no multi-gpu on Qualcomm Windows devices.

Do we intend to extend the use case beyond windows on arm for mf encoders?

nyanmisaka:
> The performance characteristics is quite strange on Qualcomm VPU as the HEVC decoding performance does not look consistent to me, it looks like, depending on the source file, the performance can have huge difference, especially on X2 Elite.
> 
> I have no idea why HDR10 and Dolby Vision HEVC sources has this huge fps difference.

Even without using any filters and encoders, are there inconsistent decoding speeds on some HEVC clips?

> For Qualcomm devices this is not a typical use case though, there is no multi-gpu on Qualcomm Windows devices.
> Do we intend to extend the use case beyond windows on arm for mf encoders?

I plan to add basic probing of the MF encoder to ffprobe. Therefore, I expect the encoder's internal behavior to be consistent, rather than selecting the first MF hw implementation enumerated.

Alternatively, let ffprobe only report the most basic D3D11VA information, including the device index. Then let the server call ffmpeg to test if the MF encoder is available. Therefore, to avoid ffmpeg crashing, either perform the test only on the winarm64 platforms or use a patch. We never know what strange platforms and drivers users might install Jellyfin on. Therefore, in my opinion, patching is a cleaner approach.

gnattu:
> Even without using any filters and encoders, are there inconsistent decoding speeds on some HEVC clips?

No. I think the reason is due to VPU resources competing as Qualcomm VPU resources are shared between decoding and encoding, without encoding the performance is much more consistent.

In decoding only test HEVC has very consistent behavior and AV1 decoding clearly becomes more resources intensive. It is so intensive that, the scale_d3d11 slows down decoding even more than our tonemap filter. 

AV1 performed better for av1 input + h264 output probably also because resources competing is less for this pipeline.

On X2 Elite, all tests still use forced output copy if using compute shader:

| Source | Baseline fps | + Filter | Filter fps | Filter cost (fps drop) | Filter cost (%) |
|--------|-------------:|----------|-----------:|-----------------------:|----------------:|
| HEVC 10-bit SDR | 299 | scale_d3d11 | 293 | 6 | 2% |
| AV1 10-bit SDR  | 257 | scale_d3d11 | 201 | 56 | 22% |
| HDR10 HEVC      | 301 | tonemap_d3d11 | 288 | 13 | 4% |
| HDR10 AV1       | 263 | tonemap_d3d11 | 252 | 11 | 4% |
| DV P5           | 286 | tonemap_d3d11 | 258 | 28 | 10% |
| HEVC 8-bit SDR  | 296 | overlay_d3d11 (ASS) | 157 | 139 | 47% |
| HEVC 8-bit SDR  | 296 | overlay_d3d11 (SRT) | 164 | 132 | 45% |

nyanmisaka:
> No. I think the reason is due to VPU resources competing as Qualcomm VPU resources are shared between decoding and encoding, without encoding the performance is much more consistent.

That makes sense.

But it's still not enough to explain the perf difference between HDR10 and DV P5 tonemapping transcoding. Perhaps DV reshape puts the GPU into a higher clock state, thus improving perf.

gnattu:
> But it's still not enough to explain the perf difference between HDR10 and DV P5 tonemapping transcoding. Perhaps DV reshape puts the GPU into a higher clock state, thus improving perf.

I don't think that is something we can control though, all we know is that our filters are unlikely to be the main performance bottleneck.

> I plan to add basic probing of the MF encoder to ffprobe. Therefore, I expect the encoder's internal behavior to be consistent, rather than selecting the first MF hw implementation enumerated.

So should I integrate this patch in this PR or you will add it later?

nyanmisaka:
I'll add it later, because testing MF on the x86 platform is more convenient for me.

debian/patches/0093-add-d3d11-deinterlace-filter.patch
```diff
@@ -0,0 +1,1326 @@
+Index: FFmpeg/configure
+===================================================================
+--- FFmpeg.orig/configure
++++ FFmpeg/configure
+@@ -3549,6 +3549,8 @@ ddagrab_filter_deps="d3d11va IDXGIOutput
+ gfxcapture_filter_deps="cxx17 threads d3d11va IGraphicsCaptureItemInterop __x_ABI_CWindows_CGraphics_CCapture_CIGraphicsCaptureSession3"
+ gfxcapture_filter_extralibs="-lstdc++"
+ scale_d3d11_filter_deps="d3d11va"
++bwdif_d3d11_filter_deps="d3d11va"
++yadif_d3d11_filter_deps="d3d11va"
```
nyanmisaka: It would be better if we could probe `d3dcompiler.h` in `configure`, just for robustness, since d3d11va is automatically enabled when `dxva.h` is available, and d3d11 filters will also be enabled accordingly. The author of `scale_d3d11` did not do any checks on the availability of `ID3D11VideoProcessor`, and I don't know how the merger allowed it in.

debian/patches/0093-add-d3d11-deinterlace-filter.patch
```diff
@@ -0,0 +1,1326 @@
...
++#endif
++
++
++#define DEINT_D3D11_ALG_YADIF 0
++#define DEINT_D3D11_ALG_BWDIF 1
++
++typedef HRESULT (WINAPI *D3DCompileProc)(LPCVOID, SIZE_T, LPCSTR,
++                                         const D3D_SHADER_MACRO *, ID3DInclude *,
++                                         LPCSTR, LPCSTR, UINT, UINT,
++                                         ID3D10Blob **, ID3D10Blob **);
```
nyanmisaka: Could we extract the loading of the d3dcompiler dll, the compilation of shaders, SRV/UAV helpers, and frame ensure/duplicate helpers into a common `d3d11_filter.{c,h}`? These boilerplate functions are essentially the same, especially when we only care about NV12, P010 and BGRA.

debian/patches/0095-add-scale-d3d11-p010-shader-fallback.patch
```diff
@@ -0,0 +1,857 @@
...
++
+     ///< Free the hardware device context reference
+     av_buffer_unref(&s->hw_frames_ctx_out);
+     av_buffer_unref(&s->hw_device_ctx);
+@@ -453,9 +1027,12 @@ static const AVFilterPad scale_d3d11_out
+ #define FLAGS (AV_OPT_FLAG_FILTERING_PARAM | AV_OPT_FLAG_VIDEO_PARAM)
+ 
+ static const AVOption scale_d3d11_options[] = {
++    { "w",      "Output video width",  OFFSET(w_expr), AV_OPT_TYPE_STRING, {.str = "iw"}, .flags = FLAGS },
+     { "width",  "Output video width",  OFFSET(w_expr), AV_OPT_TYPE_STRING, {.str = "iw"}, .flags = FLAGS },
```
nyanmisaka, line 527: nitpick: adding `ff_scale_adjust_dimensions()` would allow the use of syntax like `w=-2:h=1080`?

plus a backport of https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/21182

gnattu: This fix is already applied in this fallback addition, the early propagation is already removed.

nyanmisaka:
I'd like to know if QCOM's MFT encoder reports support for P010. Could you check using DXVA Checker?

Tests show that Intel and NVIDIA's HEVC and AV1 MFTs can encode P010. If QCOM supports it as well, it might be worth probing it.

gnattu:
> I'd like to know if QCOM's MFT encoder reports support for P010. Could you check using DXVA Checker?

Although Qualcomm's data sheet claims support for 10bit encoding, but with their current driver, you cannot even create p010 texture with encoder binding, which means there is no way to feed p010 frames to its mf encoder at all, not even the one directly from the decoder. Unless there are some hidden techniques that I do not know, there is currently no way to activate its 10bit encoding with ffmpeg.

nyanmisaka:
> Although Qualcomm's data sheet claims support for 10bit encoding, but with their current driver, you cannot even create p010 texture with encoder binding, which means there is no way to feed p010 frames to its mf encoder at all, not even the one directly from the decoder. Unless there are some hidden techniques that I do not know, there is currently no way to activate its 10bit encoding with ffmpeg.

[0001-avcodec-mfenc-misc-enhancements.patch](https://github.com/user-attachments/files/29387433/0001-avcodec-mfenc-misc-enhancements.patch)

You can try this patch. It works for me on both Intel and NVIDIA. It also fixes the async encoding issue on Intel. If it still fails on QCOM, then it means their driver does not impl P010.

gnattu:
Unfortunately no, still failed with this patch:

```
[hevc_mf @ 00000230407347f0] failed processing input: 80070057
[vost#0:0/hevc_mf @ 00000230407345b0] [enc:hevc_mf @ 0000023040151ed0] Error submitting video frame to the encoder
[vost#0:0/hevc_mf @ 00000230407345b0] [enc:hevc_mf @ 0000023040151ed0] Error encoding a frame: Generic error in an external library
[vost#0:0/hevc_mf @ 00000230407345b0] Task finished with error code: -542398533 (Generic error in an external library)
[vost#0:0/hevc_mf @ 00000230407345b0] Terminating thread with return code -542398533 (Generic error in an external library)
```

It does not fix the frame corruption either.

gnattu:
Even worse news:

It looks like your latest patch completely breaks jellyfin end to end use case for Qualcomm, at least for tone mapping:

```
[hevc_metadata @ 000001f7a179e550] No NAL unit found in packet.
[hevc_mf @ 000001f7911f0290] hevc_metadata bsffailed to receive output packet
[vost#0:0/hevc_mf @ 000001f7911efad0] [enc:hevc_mf @ 000001f791152190] Error submitting video frame to the encoder
[vost#0:0/hevc_mf @ 000001f7911efad0] [enc:hevc_mf @ 000001f791152190] Error encoding a frame: Invalid data found when processing input
[vost#0:0/hevc_mf @ 000001f7911efad0] Task finished with error code: -1094995529 (Invalid data found when processing input)
[vost#0:0/hevc_mf @ 000001f7911efad0] Terminating thread with return code -1094995529 (Invalid data found when processing input)
[hls @ 000001f791682200] Opening 'c:\Temp\file\data\cache\transcodes\c33ea014adf81bf8f2ecd9fd8c1d3c120.mp4' for writing
[out#0/hls @ 000001f79118bc70] video:2422KiB audio:0KiB subtitle:0KiB other streams:0KiB global headers:0KiB muxing overhead: unknown
frame=    7 fps=0.0 q=-0.0 Lsize=N/A time=00:00:00.00 bitrate=N/A speed=   0x elapsed=0:00:00.96
```

From jellyfin generated command:

```
ffmpeg.exe -analyzeduration 200M -probesize 1G -f mov,mp4,m4a,3gp,3g2,mj2 -init_hw_device d3d11va=dx11:0,SHADER=1 -filter_hw_device dx11 -hwaccel d3d11va -hwaccel_output_format d3d11 -noautorotate -threads 2 -i file:"C:\Temp\file\test-videos\Dolby Vision\Test Jellyfin 4K DV P5.mp4" -noautoscale -map_metadata -1 -map_chapters -1 -threads 0 -map 0:0 -map -0:a -map -0:s -codec:v:0 hevc_mf -hw_encoding 1 -compression_level 0 -b:v 44071725 -g:v:0 180 -keyint_min:v:0 180 -tag:v hvc1 -vf "setparams=color_primaries=bt2020:color_trc=smpte2084:colorspace=bt2020nc,tonemap_d3d11=format=nv12:p=bt709:t=bt709:m=bt709:tonemap=bt2390:peak=100:desat=0:force_output_copy=1"  -copyts -avoid_negative_ts disabled -max_muxing_queue_size 2048 -f hls -max_delay 5000000 -hls_time 3 -hls_segment_type fmp4 -hls_fmp4_init_filename "c:\Temp\file\data\cache\transcodes\c33ea014adf81bf8f2ecd9fd8c1d3c12-1.mp4" -start_number 0 -hls_segment_filename "c:\Temp\file\data\cache\transcodes\c33ea014adf81bf8f2ecd9fd8c1d3c12%d.mp4" -hls_playlist_type vod -hls_list_size 0 -hls_segment_options movflags=+frag_discont -y "c:\Temp\file\data\cache\transcodes\c33ea014adf81bf8f2ecd9fd8c1d3c12.m3u8
```

If it is useful for non-qualcomm platforms, can we add some command switch so that the Qualcomm can opt-out?

I confirmed that it originally works with this command before your patch.

nyanmisaka:
That metadata BSFs are intended for use with P010 to inject HDR10 attributes into the VUI. However, since the QCOM driver does not yet support 10bit encoding, I think it can be commented out.

`No NAL unit found in packet` may also be caused by the absence of the `*_mp4toannexb` BSFs.

debian/patches/0093-add-d3d11-deinterlace-filter.patch
```diff
@@ -0,0 +1,1296 @@
...
++        return temporal1(py(x, y - 1), py(x, y + 1),
++                         cy(x, y - 2), cy(x, y), cy(x, y + 2),
++                         cy(x, y - 1), cy(x, y + 1),
++                         ny(x, y - 2), ny(x, y), ny(x, y + 2),
++                         ny(x, y - 1), ny(x, y + 1), sp);
++    return temporal1(py(x, y - 1), py(x, y + 1),
++                     py(x, y - 2), py(x, y), py(x, y + 2),
++                     cy(x, y - 1), cy(x, y + 1),
++                     cy(x, y - 2), cy(x, y), cy(x, y + 2),
++                     ny(x, y - 1), ny(x, y + 1), sp);
```
nyanmisaka: Perhaps check if the d3d11 deint filters are also affected by these typos (https://github.com/jellyfin/jellyfin-ffmpeg/commit/165040276090fdde4ac8551d07a4b2e744849c06)? They were all ported from the cuda filters.

gnattu: Yes it is affected

nyanmisaka (approved):

PR #742, opened 2026-07-15
Fix rotated HDR10 mastering-display primaries in nvenc HEVC

lou-perret:
Hello!

I ran into an issue with HDR10 movies: after encoding with hevc_nvenc, the picture came out washed-out / too bright.

I used AI to help me (I'm not an ffmpeg pro) and managed to diagnose that the mastering-display color primaries were getting rotated (red shifted to green, green shifted to blue, blue shifted to red).

I found that 0032-add-hdr-metadata-for-nvenc-hevc-encoder.patch uses {2, 0, 1} for the mapping, but HEVC stores the primaries in green, blue, red order. Since FFmpeg's display_primaries array is in RGB order (R=0, G=1, B=2), the mapping that produces G, B, R should be {1, 2, 0}, not {2, 0, 1} (the comment right above the line even says "HEVC uses a g,b,r ordering, which we convert from a more natural r,g,b").

After this change, it seems to fix the issue, but again, I'm not an ffmpeg pro, so I may be mistaken.

Thank you :)

### DIFF
--- a/debian/patches/0032-add-hdr-metadata-for-nvenc-hevc-encoder.patch
+++ b/debian/patches/0032-add-hdr-metadata-for-nvenc-hevc-encoder.patch
@@ -23,7 +23,7 @@ Index: FFmpeg/libavcodec/nvenc.c
 +        if (sd) {
 +            AVMasteringDisplayMetadata *mdm = (AVMasteringDisplayMetadata *)sd->data;
 +            // HEVC uses a g,b,r ordering, which we convert from a more natural r,g,b
-+            const int mapping[3] = {2, 0, 1};
++            const int mapping[3] = {1, 2, 0};
 +            const int chroma_den = 50000;
 +            const int luma_den = 10000;
 +


gnattu (changes requested):
LLM understands the intention wrong. We are converting gbr back to rgb, not from rgb to gbr and llm did the wrong change

gnattu:
> the picture came out washed-out / too bright

You need to do tone mapping if you are converting hdr to sdr. We already provided tonemap_cuda filter that you can use.

nyanmisaka:
@gnattu https://github.com/search?q=repo%3Ajellyfin%2Fjellyfin-ffmpeg%20%22const%20int%20mapping%5B3%5D%22&type=code

I’ve verified the fix, and it is correct. The metadata channels for the decoder and encoder differ. This patch originates from an old patch, where the author likely copied the channel order directly from the decoder.

While newer versions of the ffnvcodec have added native HDR metadata writing capabilities, we cannot upgrade to a higher version at the moment, so we will likely need to keep using this patch for a while longer.

nyanmisaka (approved):
HEVC (GBR) -> AV (R2 G0 B1)

AV (RGB) -> HEVC (G1 B2 R0)

gnattu (approved):

Shadowghost (approved):

PR #747, opened 2026-07-21
Fix over release in yadif_videotoolbox filter

nyanmisaka:
**Changes**
- Fix over release in yadif_videotoolbox filter
- Update more backported vulkan fixes from upstream
- Enable vulkan-static for DEB builds

**Issues**
- https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23850

### DIFF
--- a/debian/patches/0056-add-bwdif-videotoolbox-filter.patch
+++ b/debian/patches/0056-add-bwdif-videotoolbox-filter.patch
@@ -804,7 +804,17 @@ Index: FFmpeg/libavfilter/vf_yadif_videotoolbox.m
 ===================================================================
 --- FFmpeg.orig/libavfilter/vf_yadif_videotoolbox.m
 +++ FFmpeg/libavfilter/vf_yadif_videotoolbox.m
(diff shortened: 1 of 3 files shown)


gnattu (approved):
This is okay because the objc objects are automatically released anyway, but the more correct practice (for objc) is to wrap all objects that is meant to be auto released in an auto release pool to make the behavior explicit. This historical code means at least on some version on macOS/ffmpeg these two objects were not auto released (it has no implicit auto release pool). But on which version and what combination is now hard to find out.

Upstream does not like auto release objects for some reason but it is not always possible to manually manage the lifetime of objc objects and relying on auto release pool is not the "you like it or not" problem, but "it is the only way", as a lot of objc objects have autorelease internally.

For example: https://github.com/KhronosGroup/MoltenVK/pull/2436

nyanmisaka:
This is the simplest fix for our FFmpeg build. Once upstream accepts that PR, I agree that we should also switch to using `autoreleasepool` in other Metal filters where appropriate.

Shadowghost (approved):

PR #752, opened 2026-08-12
Bump version to 8.1.2-3

nyanmisaka:
**Changes**
- Update backported vulkan fixes from upstream
- Polish some DX11-OCL sharing patches
- Update fixes for relaxing safe filenames in mkv attachments
- Backport a fix to not stall sub2video on stream EOF
- Fix fate tests errors in tests/ref/fate/source
- Backport Mesa fixes for compatibility with older AMDGPU FW
- Bump version to 8.1.2-3

**Issues**
- https://github.com/jellyfin/jellyfin/pull/16293#issuecomment-3953287803

### DIFF
--- a/debian/patches/0010-add-a-hack-for-opencl-reverse-mapping.patch
+++ b/debian/patches/0010-add-a-hack-for-opencl-reverse-mapping.patch
@@ -66,7 +66,7 @@ Index: FFmpeg/libavfilter/vf_hwmap.c
  #include "libavutil/opt.h"
  #include "libavutil/pixdesc.h"
  
-+#if HAVE_OPENCL_D3D11
++#if CONFIG_D3D11VA
 +#include "libavutil/hwcontext_d3d11va.h"
 +#endif
 +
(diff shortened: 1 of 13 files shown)


debian/patches/0010-add-a-hack-for-opencl-reverse-mapping.patch
```diff
@@ -111,9 +101,11 @@ Index: FFmpeg/libavfilter/vf_hwmap.c
 +                frames->initial_pool_size += (avctx->extra_hw_frames > 2 ? avctx->extra_hw_frames : 2);
 +            }
 +
-+#if HAVE_OPENCL_D3D11
-+            if (frames->format == AV_PIX_FMT_D3D11)
-+                frames->user_opaque = &texDesc;
++#if CONFIG_D3D11VA
++            if (frames->format == AV_PIX_FMT_D3D11) {
++                AVD3D11VAFramesContext *frames_d3d11 = frames->hwctx;
++                frames_d3d11->BindFlags = D3D11_BIND_DECODER;
```
gnattu, line 107: This  makes the old `|=` to `=` and may break direct texture reading from d3d11va decoder to d3d11 filters, at least it was the case when I was developing the d3d11 filters. Can you check the compatibilities?

nyanmisaka: This path is intended solely for DX11-OpenCL **reverse** mapping. Could you elaborate on how it interacts with pure D3D11VA and D3D11 filters pipelines?

Or are you referring to the one in hwupload?

debian/patches/0010-add-a-hack-for-opencl-reverse-mapping.patch
```diff
@@ -168,26 +152,16 @@ Index: FFmpeg/libavutil/hwcontext_d3d11va.c
...
          .BindFlags  = hwctx->BindFlags,
 -        .MiscFlags  = hwctx->MiscFlags,
 +        .MiscFlags  = hwctx->MiscFlags | D3D11_RESOURCE_MISC_SHARED,
      };
  
-+#if HAVE_OPENCL_D3D11
-+    if (ctx->user_opaque) {
-+        D3D11_TEXTURE2D_DESC *desc = ctx->user_opaque;
-+        if (desc->BindFlags & D3D11_BIND_DECODER)
-+            texDesc.BindFlags |= D3D11_BIND_DECODER;
```
gnattu, line 183: I am talking about the case moved from there. This `|=` was explicitly added so that the old flags are not cleared and d3d11 filters can read the texture.

nyanmisaka: I realize that. We no longer misuse `user_opaque`.

I am wondering where the **old flags** come from, given that the `AVD3D11VAFramesContext` in `hwmap` and `hwupload` is newly created. The `hwcontext_d3d11va.c` allocator does not currently set any BindFlags by default.

gnattu (approved):

Shadowghost (approved):

PR #756, opened 2026-09-05
Drop support for EOL Debian Bullseye (old old stable)

nyanmisaka:
The dependencies required to build it are no longer available.

**Changes**
- Drop support for EOL Debian Bullseye (old old stable)

### DIFF
--- a/.github/workflows/_meta.yaml
+++ b/.github/workflows/_meta.yaml
@@ -8,7 +8,7 @@ on:
       codenames:
         description: 'Stringified JSON object listing target distro codenames'
         required: true
-        default: '["bullseye"]'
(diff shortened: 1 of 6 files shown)


Shadowghost (approved):

.github/workflows/_meta.yaml
```diff
@@ -8,7 +8,7 @@ on:
       codenames:
         description: 'Stringified JSON object listing target distro codenames'
         required: true
-        default: '["bullseye"]'
+        default: '["bookworm"]'
```
Shadowghost: Should we directly go to trixie?

gnattu (approved):

PR #757, opened 2026-09-06
Make HDR10+ removal more precise

gnattu:
Actually parse SEI messages to avoid removing unrelated messages in NAL. This may help with some fragile players that would crash the playback after hdr10+ removal for HEVC.

The performance difference is not observable on my system.

**Changes**

**Issues**


Shadowghost (approved):

nyanmisaka (approved):
I don't quite understand why deleting the optional SEI causes the decoder to crash. But it's good to avoid unnecessary deletions.

gnattu:
> I don't quite understand why deleting the optional SEI causes the decoder to crash. But it's good to avoid unnecessary deletions.

It is just a speculation because some users report even manually stripped hdr10+ source caused a full transcode, and the very likely reason is that the original playback just crashed and the fallback playback forced a full transcode.

PR #766, opened 2026-09-12
Backport trim bitstream filter

gnattu:
backported from https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/23839

**Changes**

This backports upstream newly implemented `trim` bsf that can also handle the problem we have in:
https://github.com/jellyfin/jellyfin/pull/16580, and replaces the `noise` bsf used in that PR.

**Issues**

### DIFF
--- a/debian/patches/0100-backport-trim-bitstream-filter.patch
+++ b/debian/patches/0100-backport-trim-bitstream-filter.patch
@@ -1,0 +1,1010 @@
+Index: FFmpeg/doc/bitstream_filters.texi
+===================================================================
+--- FFmpeg.orig/doc/bitstream_filters.texi
++++ FFmpeg/doc/bitstream_filters.texi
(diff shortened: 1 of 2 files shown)


nyanmisaka (commented):
Does it offer any advantages over the noise BSF?

Also, the `patches/series` file is missing.

gnattu:
> Does it offer any advantages over the noise BSF?

The upstream PR has a reason:
> Its functionalities cover some of the noise filter's functionalities but the purpose is totally different and noise would be quite a confusing filter to use for that.

And this filter can do trim inside a packet instead of dropping the packet entirely like we are doing with the noise filter currently.

nyanmisaka (approved):
So, it allows for finer-grained modifications, which justifies its existence.

Shadowghost (approved):

